PR #210, opened 2013-08-14
Improved some translations, made the code more PEP8

bcho:


zh-cn/python-cn.html.markdown
```diff
@@ -116,81 +117,81 @@ None is None  #=> True
...
-li[:3] #=> [1, 2, 4]
+li[:3]  # => [1, 2, 4]
 
 # 删除特定元素
-del li[2] # li 现在是 [1, 2, 3]
+del li[2]  # li 现在是 [1, 2, 3]
 
 # 合并列表
-li + other_li #=> [1, 2, 3, 4, 5, 6] - 不改变这两个列表
+li + other_li  # => [1, 2, 3, 4, 5, 6] - 并不会不改变这两个列表
```
but0n, line 166: It becomes a double negative sentence here, the meaning is totally opposite.
It should be "并不会改变这两个列表"

bcho: Nice catch! Could you send a pull request for that :)

PR #553, opened 2014-03-01
[json/en] Added JSON.

### LINKED ISSUE #551: [Request] JSON
As one of the great alternatives to XML, it would be great to have something about JSON. Since I'm working on [XML](https://github.com/adambard/learnxinyminutes-docs/issues/550) and don't know much about this format, I'd like to see a good reference to JSON on LXIYM.

iirelu:
The entire specification is shown here, for the most part. I had plenty of room to work with so I threw in some jokes.

Resolves #551


aloisdg:
I translated your document from en to fr here : #1078

Nice job !

PR #851, opened 2014-11-10
Update java-cn.html.markdown

erc1IC:
输出错误


geoffliu:
@adambard I'm fixing this part in #931, had some free time on my hand. :)

PR #890, opened 2014-12-20
[scala] Fixed function return type

arturmkrtchyan:
Scala compiler fails to infer the return type of the function when there's an explicit return statement used in the function

### DIFF
--- a/scala.html.markdown
+++ b/scala.html.markdown
@@ -199,7 +199,7 @@ weirdSum(2, 4)  // => 16
 
 // The return keyword exists in Scala, but it only returns from the inner-most
 // def that surrounds it. It has no effect on anonymous functions. For example:
-def foo(x: Int) = {
+def foo(x: Int): Int = {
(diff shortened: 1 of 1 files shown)


arturmkrtchyan:
Additionally can be beneficial if you know how to use it.

Like implementing indexOf
```scala
def indexOf(str: String, ch: Char): Int = {
  var i = 0
  until (i == str.length) {
    if (str(i) == ch) return i
    i += 1
  }
  return -1
}
```  

Any chance to merge it ?  I think at least in tutorial the syntax part should be correct. 

Now scala compiler is not able to compile that sample.

levibostian:
I think we are ok with merging it, we just want more of a note to the user that this is a bad style to do in general. You fixing the compilation error is great, would you please also add a note in there for us so we can merge that in with it?

arturmkrtchyan:
@levibostian warning note is added to avoid return statement.

adambard:
Great, thanks!

PR #894, opened 2014-12-22
[julia/ja] Translating X=Julia into Japanese

yomichi:


adambard:
Sorry we didn't merge this sooner, thanks a lot!

PR #898, opened 2014-12-23
[YAML/en] translated to [YAML/de]

switchhax:
First translation, which i have to offer. Hope with that everything is all right with the structure of the document.


adambard:
Thanks!

switchhax:
I´ll work on other translations as soon as the exam-time in German schools end.

PR #903, opened 2014-12-26
[commonlisp/pt] - Translate Common Lisp guide to Portuguese Brazil

edipofederle:


adambard:
Probably fine, thanks!

PR #912, opened 2015-01-07
java: Add `@Override` for interfaces.

weakish:
It is a good practice to use `@Override` when implementing methods
for interfaces.


vendethiel:
Thanks!

PR #915, opened 2015-01-08
C: two minor fixes

mordner:
Fixed one line being cut off and a comment about the export attribute was referring to the wrong function and variable.


levibostian:
Thanks!

PR #917, opened 2015-01-11
[en/Scala] Rewrite the pattern matching section

geoffliu:


scala.html.markdown
```diff
@@ -407,41 +407,55 @@ val otherGeorge = george.copy(phoneNumber = "9876")
 // 6. Pattern Matching
 /////////////////////////////////////////////////
 
-val me = Person("George", "1234")
+// Pattern matching is a powerful and commonly used feature in Scala. Here's how
+// you pattern match a case class. NB: Unlike other languages, Scala cases do
+// not have breaks.
```
levibostian: If Scala doesn't have breaks, what do you do instead?

geoffliu: Ah, poor wording. Scala doesn't need breaks between cases, fall-through doesn't happen. I'll update.

levibostian:
Thanks!

scala.html.markdown
levibostian: Ok thanks. Go ahead and make another PR when you have it complete.

PR #924, opened 2015-01-13
[C++/en] translated to [C++/pt-br]

miguelarauj1o:


adambard:
If anyone wants to edit, they can do it live!

Thanks!

PR #926, opened 2015-01-15
[rust/en] Update the Rust tutorial

### LINKED ISSUE #860: Rust tutorial should discuss type inference more.
At present, Rust's type inference only gets a very brief mention. I don't know Rust very well, but most of the example code on the Rust website uses type inference rather than explicit type annotations, so it would appear that this is the idiomatic thing. 

Another issue with defaulting to explicit type annotations is that people who are first introduced to the language via Learn X in Y Minutes may end up assuming that explicit type annotations are compulsory in Rust, and there are a lot of programmers out there who would be put off by that.

P1start:
This adjusts the English Rust tutorial for changes to the language and generally tweaks a few other things.

Fixes #860.

### DIFF
--- a/rust.html.markdown
+++ b/rust.html.markdown
@@ -7,9 +7,13 @@ filename: learnrust.rs
 
 Rust is an in-development programming language developed by Mozilla Research.
 It is relatively unique among systems languages in that it can assert memory
-safety *at compile time*. Rust’s first alpha release occurred in January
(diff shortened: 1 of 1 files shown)


adambard:
Looks good to me (Although I'm suspicious of the changed en-dashes, if those break encoding I'll be coming after you)

Thanks!

PR #935, opened 2015-01-20
Explain Optional Chaining

keito:
Without going into too much detail, mention optional chaining so it's
not confusing that there is a question mark suffix (it's not currently
mentioned anywhere on the page).


adambard:
Sounds good, thanks!

keito:
Thanks! :)

PR #940, opened 2015-01-28
[swift/en,cn,ru]Updating the getting started guide weblink.

TheDmitry:


geoffliu:
Could update the file in pt-br as well? Thanks!

TheDmitry:
Commited.
But pt-br is not visible on the LearnXinYMinutes homepage. Is it broken?

geoffliu:
Looks like there's an extra comma in there. I'll fix it.

geoffliu:
Oh the pt-br version has an extra comma in there, so the header section isn't parsing. I'll fix it. Thanks!

PR #941, opened 2015-01-28
[python/en] Add some cool Python learning sources

armab:
[Codecademy](http://www.codecademy.com/en/tracks/python) has free interactive course great for beginners,
while http://www.checkio.org/ is online game to learn Python while playing and solving problems.

Both are absolutely cool way to learn the Python in interesting and interactive way

### DIFF
--- a/python.html.markdown
+++ b/python.html.markdown
@@ -614,7 +614,6 @@ print(say(say_please=True))  # Can you buy me a beer? Please! I am poor :(
 
 ### Free Online
 
-* [CodeCademy interactive Course for Beginners](http://www.codecademy.com/en/tracks/python)
 * [Learn Python The Hard Way](http://learnpythonthehardway.org/book/)
 * [Dive Into Python](http://www.diveintopython.net/)
 * [The Official Docs](http://docs.python.org/2.6/)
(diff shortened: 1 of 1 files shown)


levibostian:
Not sure if we typically allow online resources or just free books. Will have to see what others think about this one. 

Thanks!

geoffliu:
I'm quite a fan of Codecademy as a beginners' tool, but LXIY is targeted towards people who already know programming, and just need a quick course in a new language. I suspect people who read our pages will find Codecademy too slow, but I'm not opposed to adding it.

checkio I tried a while ago, and it was buggy back then. I just tried it again, and I was thrown into a redirect loop. Maybe we can reconsider adding them if they become stable, but right now doesn't seem to be the right time.

armab:
I see, just removed CodeCademy link.

Regarding CheckIO - here is Guido van Rossum's profile: http://www.checkio.org/user/guido/
Also he participating in code reviews from time to time: http://www.checkio.org/blog/guido-van-rossums-code-review/
So I think It's good enough resource and worth adding.

Actually I never had any issues like redirect loop or something like that in CheckIO, but you can [fill bug report](https://www.checkio.org/forum/add/?source=footer&q=tag%3Afeedback) if have time.

Thanks.

PR #943, opened 2015-01-29
Update scala.html.markdown

ikr0m:
Removed typo (unnecessary double quote in patternFunc)


geoffliu:
Oops! Thanks for the correction!

PR #944, opened 2015-01-29
[r/en] Add definition of parakeet variable

acganesh:
Added a definition of the `parakeet` variable.  It's currently never defined, but referenced.


levibostian:
Thanks!

PR #947, opened 2015-01-31
[elixir/en] Add ranges under 'Basic types'

microamp:
The list's getting bigger, but figured it's worth mentioning ranges since it's one of the value types. Thanks for reviewing.

(cc/ @mrshankly and @dskecse)


levibostian:
Thanks!

PR #948, opened 2015-01-31
[CSharp/en]Some fixes and additions

Laoujin:
I'd like to add and restructure more but let's first find out if this is acceptable as pull request :)


csharp.html.markdown
```diff
@@ -33,7 +36,7 @@ using System.Threading.Tasks;
 using System.IO;
 
 // defines scope to organize code into "packages"
-namespace Learning
+namespace LearningXInYMinutes.CSharp
```
levibostian: You included a '.' character here. Does that change anything at all? Does that need a little description to explain what it does?

csharp.html.markdown
```diff
@@ -18,13 +19,15 @@ C# is an elegant and type-safe object-oriented language that enables developers
...
+/// This is an XML documentation comment which can be used to generate external
+/// documentation or provide context help within an IDE
 /// </summary>
+//public void MethodOrClassOrOtherWithParsableHelp() {}
 
 // Specify namespaces application will be using
 using System;
 using System.Collections.Generic;
-using System.Data.Entity;
+using System.Data.Entity; // Add dll reference with NuGet: Install-Package EntityFramework
```
levibostian: Maybe instead of having a comment that explains what this one statement does, have a description above all of them explaining what they all do.

levibostian:
Besides the couple of things I commented on, this looks good! Thanks!

csharp.html.markdown
Laoujin: All other `usings` are available right away but System.Data.Entity needs a download first. So if someone loaded this in Visual Studio, there would be a compile error. That is until one does "Install-Package EntityFramework" in the NuGet Console.

Perhaps I should add a tidbit about NuGet? (=pip/npm/... for .NET)
Or explain why `System.Data.Entity` is not available out-of-the-box?

csharp.html.markdown
Laoujin: I wanted to show how multiple levels of nesting work for "packages".

Should I add a little bit about namespace scoping?

levibostian: Yes please. Just a little note about it.

csharp.html.markdown
levibostian: Yes please. After you explained that I understood what was going on but until then, I didn't understand this.

Laoujin:
How about this? :)

levibostian:
So close @Laoujin. After that quick little typo, this will get merged.

Laoujin:
Oops! Fixed :)

PR #949, opened 2015-02-01
[obj-c/en] Fix typos in objective-c doc

karlwithak:
One code and four comment fixes.

### DIFF
--- a/objective-c.html.markdown
+++ b/objective-c.html.markdown
@@ -55,7 +55,7 @@ int main (int argc, const char * argv[])
     id       myObject2 = nil;  // Weak typing
     // %@ is an object
     // 'description' is a convention to display the value of the Objects
-    NSLog(@"%@ and %@", myObject1, [myObject2 description]); // Print "(null) and (null)"
+    NSLog(@"%@ and %@", myObject1, [myObject2 description]); // prints "(null) and (null)"
     
     // String
     NSString *worldString = @"World";
@@ -70,7 +70,7 @@ int main (int argc, const char * argv[])
     char theLetterZ            = [theLetterZNumber charValue]; // or 'Z'
     NSLog(@"%c", theLetterZ);
 
-    // Integral literals
+    // Integer literals
     NSNumber *fortyTwoNumber = @42;
     int fortyTwo             = [fortyTwoNumber intValue]; // or 42
     NSLog(@"%i", fortyTwo);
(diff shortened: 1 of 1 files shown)


objective-c.html.markdown
```diff
@@ -70,7 +70,7 @@ int main (int argc, const char * argv[])
     char theLetterZ            = [theLetterZNumber charValue]; // or 'Z'
     NSLog(@"%c", theLetterZ);
 
-    // Integral literals
```
geoffliu: For this line, I like the old wording better. "Integer" would be ambiguous here, since it can take the mathematical sense (and indeed all the literals below are integers), or the programming sense of `int`.

objective-c.html.markdown
```diff
@@ -128,9 +128,10 @@ int main (int argc, const char * argv[])
     // May contain different data types, but must be an Objective-C object
     NSArray *anArray      = @[@1, @2, @3, @4];
     NSNumber *thirdNumber = anArray[2];
-    NSLog(@"Third number = %@", thirdNumber); // Print "Third number = 3"
-    // NSMutableArray is mutable version of NSArray allowing to change items in array
-    // and extend or shrink array object. Convenient, but not as efficient as NSArray
+    NSLog(@"Third number = %@", thirdNumber); // prints "Third number = 3"
+    // NSMutableArray is a mutable version of NSArray allowing you to change
```
geoffliu: While we're at it:

NSMutableArray is a mutable version of NSArray, allowing you to change the items in the array and to extend or shrink the array object. Convenient, but not as efficient as NSArray.

objective-c.html.markdown
```diff
@@ -55,7 +55,7 @@ int main (int argc, const char * argv[])
     id       myObject2 = nil;  // Weak typing
     // %@ is an object
     // 'description' is a convention to display the value of the Objects
-    NSLog(@"%@ and %@", myObject1, [myObject2 description]); // Print "(null) and (null)"
+    NSLog(@"%@ and %@", myObject1, [myObject2 description]); // prints "(null) and (null)"
```
geoffliu: Nit: why lowercase?

geoffliu:
@karlwithak Thanks for the PR, and sorry to be so nit picky. I feel that if we're fixing minor mistakes here, might as well go all the way. :)

karlwithak:
I agree with you about Integral vs. Integer, not sure what I was thinking.
I changed to lower-case 'print' because it was used everywhere else. I then realized I might as well change all print comments to: print => "printed value" to be completely consistent.

levibostian:
Thanks!

Issue #526, opened 2014-02-13
[Request] Computer Science topics outside of programming


gizmo385:
I think the addition of pages on things related to programming and computer science, but not purely programming would be helpful. For example, a page covering sorting algorithms and their differences or a page on basic algorithm analysis.

Thoughts?

Jakehp:
@levibostian @adambard I started this today. I am thinking we should create a new section for "Algorithms & Data Structures". What do you think?

levibostian:
@Jakehp you have my vote!

Jakehp:
https://github.com/adambard/learnxinyminutes-docs/pull/950

gizmo385:
Definite (albeit late) +1 on the contribution by @Jakehp. Is there any way we'd be able to use MathJax or something similar to parse LaTeX? Once the explanations start delving deeper into some of the more complicated math related to algorithms and data structures, certain things like sums are going to become commonplace and we'll need a way to effectively express those to the users. Thoughts?

Jakehp:
@gizmo385 That's a really good idea; although I'll have to learn LaTeX. @adambard is this possible?

levibostian:
I've experienced MathJax.js in the past and I really liked it. Just a simple [js file](https://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS-MML_HTMLorMML) you stick in the head and anywhere in your HTML you enclose Latex. 

Here is a fully working example:  
```
<!DOCTYPE html>
<html>
<head>
<title>MathJax TeX Test Page</title>
<script type="text/x-mathjax-config">
  MathJax.Hub.Config({tex2jax: {inlineMath: [['$','$'], ['\\(','\\)']]}});
</script>
<script type="text/javascript"
  src="http://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS-MML_HTMLorMML">
</script>
</head>
<body>
When $a \ne 0$, there are two solutions to \(ax^2 + bx + c = 0\) and they are
$$x = {-b \pm \sqrt{b^2-4ac} \over 2a}.$$
</body>
</html>
```

So really, we could do that pretty easy.

geoffliu:
How would that play with other pages? Would that screw up, for example, the Perl page?

Can we configure that somehow to use `<latex>` tags instead of `$ $`?

geoffliu:
Well, using `$ $` is a bit dangerous, especially for the Perl page, no?

[...]

PR #950, opened 2015-02-01
asymptotic notation rough draft

Jakehp:
I gave this a re-read through but would definitely appreciate several more sets of eyes reading over it....

Is this clear?
Not enough explanation? Too much?
I worry this is too short to actually grasp the concept, but I think it's enough to get a basic understanding?

Once this gets cleared/merged I plan on writing up a doc on Analyzing Algorithms (big oh notation application https://github.com/adambard/learnxinyminutes-docs/issues/526#issuecomment-41224568).

### DIFF
--- a/asymptotic-notation.html.markdown
+++ b/asymptotic-notation.html.markdown
@@ -38,8 +38,9 @@ You can label a function, or algorithm, with an Asymptotic Notation in many diff
 Some examples are, you can describe an algorithm by it's best case, worse case, or equivalent case.
 The most common is to analyze an algorithm by it's worst case. This is because if you determine an
 algorithm's run time or time complexity, by it's best case, what if it's best case is only obtained
-given at a low, unrealistic, input size? It is equivalent to having a 5 meter sprinting race.
-That isn't the best measurement.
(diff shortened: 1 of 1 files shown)


levibostian:
Well done @Jakehp! 

I am not quite sure how it is going to look as far as the site goes because this is a new section. The best way in my opinion to see is to merge it and see what it looks like and fix from there. So here it goes...

Jakehp:
Welllllllp
http://learnxinyminutes.com
http://learnxinyminutes.com/docs/asymptotic-notation/

vendethiel:
I don't think the website builder is happy with so little metadata ;-).

Jakehp:
So what's the plan on fixing this?

vendethiel:
waiting for @adambard .. :)

adambard:
Right, on it.

adambard:
Ok, so, now you can add new categories by naming them as in this PR, but you need to add a `name` field (which I've done for this article). This is the same as the `language` field in languages or the `tool` field in tools, and is clearly better, but there's no way I'm going back and changing all those over individually :P

Jakehp:
Sounds good, thanks!

PR #951, opened 2015-02-01
Add more information on ranges in Haskell

devinmcginty:


haskell.html.markdown
```diff
@@ -67,10 +67,18 @@ not False -- True
 ----------------------------------------------------
 
 -- Every element in a list must have the same type.
--- Two lists that are the same
+-- These two lists are the same
```
vendethiel: do you mind adding a colon? :)

haskell.html.markdown
```diff
@@ -67,10 +67,18 @@ not False -- True
...
 [1, 2, 3, 4, 5]
 [1..5]
 
+-- Ranges are versatile.
+['A'..'F'] -- "ABCDEF"
```
vendethiel: is the fact that String == [] Char` explained before?

devinmcginty: It was mentioned, but somewhat easy to miss. I added a second example of a string as a list.

vendethiel:
Thanks!

PR #954, opened 2015-02-03
javascript/en: added some md formatting

uvtc:
Added some markdown formatting to the comments throughout to clarify meaning. Also tweaked a couple of other quote marks as well. Also shortened a couple of long lines.

More to do, but stopped here for now. :)

### DIFF
--- a/javascript.html.markdown
+++ b/javascript.html.markdown
@@ -110,19 +110,19 @@ null === undefined; // = false
 13 + !0; // 14
 "13" + !0; // '13true'
 
-// You can access characters in a string with charAt
+// You can access characters in a string with `charAt`
 "This is a string".charAt(0);  // = 'T'
 
-// ...or use substring to get larger pieces
+// ...or use `substring` to get larger pieces.
 "Hello world".substring(0, 5); // = "Hello"
 
-// length is a property, so don't use ()
+// `length` is a property, so don't use ().
 "Hello".length; // = 5
 
-// There's also null and undefined
-null; // used to indicate a deliberate non-value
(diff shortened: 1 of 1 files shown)


geoffliu:
This is very helpful to readability, thanks!

I wonder how others feel about this issue, but the lines you're editing are in the code section of the page. The backticks don't actually get converted to formatting, they would just show up as backticks. I've always thought single quotes in comments are less jarring, but that could just be me.

levibostian:
...and me. I agree with you @geoffliu

adambard:
IMO this passes the better-than-before test, so I'm just going to merge it and we'll see if anyone cares strongly enough to correct it :)

uvtc:
@geoffliu , yes, certainly, the markdown formatting I've added is not (currently) being processed by a markdown implementation, but I think it provides context and adds readability for humans. For example, "the this keyword" vs "the `this` keyword" --- the backticks make it clear we're talking about the "this" in the code, and that it's not just a typo. An even better example: "the new keyword" (*"a more recent one?"*) vs. "the `new` keyword".

As an aside, it's also becoming more common to format code comments and docstrings in markdown so that doc generators can (now or down the road) extract those and generate nicely-rendered html from them (A few I'm aware of: Rust does this, Julia will be doing so in their next release (v0.4), and Clojure is has a well-known third-party tool ([Marginalia](https://github.com/gdeer81/marginalia)) for it).

In the places I changed single quotes do double quotes --- sorry, that was not markdown formatting but simply a minor typographic correction which maybe I should have kept for a separate PR.

PR #960, opened 2015-02-10
minor fixes, adding color echo output

hzlmn:
Fix some shellcheck common mistakes and add echo color output

### DIFF
--- a/bash.html.markdown
+++ b/bash.html.markdown
@@ -32,14 +32,14 @@ echo "Hello world!"
 echo -e "\033[0;32mHello world!\033[0m"
 
 # You can use these codes
-Black        0;30     Dark Gray     1;30
-Blue         0;34     Light Blue    1;34
-Green        0;32     Light Green   1;32
-Cyan         0;36     Light Cyan    1;36
-Red          0;31     Light Red     1;31
-Purple       0;35     Light Purple  1;35
-Brown/Orange 0;33     Yellow        1;33
-Light Gray   0;37     White         1;37
+# Black        0;30     Dark Gray     1;30
+# Blue         0;34     Light Blue    1;34
+# Green        0;32     Light Green   1;32
+# Cyan         0;36     Light Cyan    1;36
+# Red          0;31     Light Red     1;31
+# Purple       0;35     Light Purple  1;35
+# Brown/Orange 0;33     Yellow        1;33
+# Light Gray   0;37     White         1;37
 
 # Each command starts on a new line, or after semicolon:
 echo 'This is the first line'; echo 'This is the second line'
--- a/bash.html.markdown
+++ b/bash.html.markdown
@@ -25,7 +25,21 @@ Nearly all examples below can be a part of a shell script or executed directly i
 # As you already figured, comments start with #. Shebang is also a comment.
 
 # Simple hello world example:
-echo Hello world!
+echo "Hello world!"
+
+# Color output of echo command
+# This command will show green 'Hello world!' text
+echo -e "\033[0;32mHello world!\033[0m"
+
+# You can use these codes
+Black        0;30     Dark Gray     1;30
+Blue         0;34     Light Blue    1;34
+Green        0;32     Light Green   1;32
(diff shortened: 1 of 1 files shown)


bash.html.markdown
```diff
@@ -25,7 +25,21 @@ Nearly all examples below can be a part of a shell script or executed directly i
 # As you already figured, comments start with #. Shebang is also a comment.
 
 # Simple hello world example:
-echo Hello world!
+echo "Hello world!"
```
geoffliu: Careful here. The `!"` at the end will be interpreted as history substitution. Use single quotes here, or I would just leave the line as is, since it works as intended.

bash.html.markdown
```diff
@@ -25,7 +25,21 @@ Nearly all examples below can be a part of a shell script or executed directly i
...
+
+# Color output of echo command
+# This command will show green 'Hello world!' text
+echo -e "\033[0;32mHello world!\033[0m"
```
geoffliu: Same thing here, `!\033[0m` will be interpreted as history substitution.

bash.html.markdown
```diff
@@ -25,7 +25,21 @@ Nearly all examples below can be a part of a shell script or executed directly i
...
+
+# You can use these codes
+Black        0;30     Dark Gray     1;30
```
geoffliu: Could you comment out these lines?

bash.html.markdown
```diff
@@ -69,12 +83,11 @@ echo "Scripts arguments seperated in different variables: $1 $2..."
 # Reading a value from input:
 echo "What's your name?"
 read NAME # Note that we didn't need to declare a new variable
-echo Hello, $NAME!
+echo "Hello, $NAME!"
```
geoffliu: Again, here.

bash.html.markdown
drKraken: If i correctly understand you, `!!` used for such purpose.

bash.html.markdown
drKraken: This command will show color output of echo command. You can use different codes for colors 0;31 -red 0;30 black  etc.

bash.html.markdown
geoffliu:
```
bash-3.2$ echo "Hello world!"
bash: !": event not found
```

drKraken:
```
lehjs@LenovoPC:~$ echo "Hello world!"
echo "Hello world!"
Hello world!
```
Hm, it interps correct on my bash 4.3.30

geoffliu: Ah, must be the version difference. The thing is, Bash 3.2 is still the one bundled on Macs, and there's a significant developer population using Macs. Does it work in 4.3 if you escape the `!` with a backslash? As in

`echo "Hello world\!"`

bash.html.markdown
geoffliu: Right but `Black` isn't a command on most machines, nor are all the other color names. It's nice to have the color codes listed here, but they should be as comments.

bash.html.markdown
drKraken: @geoffliu see c0384133d6a5b3b765d9ae5ef22212105a41474c

bash.html.markdown
vendethiel: Please remove the example :).

drKraken: which one?

vendethiel: I mean, remove the "!" that work for you but not in older bashes.

deryni:
Sequences for color codes should not be included as literals. They aren't guaranteed to be the same everywhere. Use tools like `tput` to get the correct sequence for the *current* terminal.

deryni:
History expansion is not at all smart. As such it should either be explicitly disabled with `set +H` or `!` should be entirely avoided.

vendethiel:
@olehkuchuk ?

bash.html.markdown
vendethiel: @hzlmn ping?:)

PR #961, opened 2015-02-12
[ocaml/en] Change the example of mutually recursive functions

duanchenggang:
Change the example to a meaningful and self-contained example of mutually recursive functions.

And fix a typo.


vendethiel:
Thanks! :+1:

PR #965, opened 2015-02-18
[hack/en] Add english Hack language page

stecman:
[The Hack language](http://hacklang.org/)

This article covers the major features of the Hack language. A few details are currently missing, but these aren't critical for understanding what the language offers. (Missing items include collections, `invariant()`, and predefined Interfaces and classes).

### DIFF
--- a/hack.html.markdown
+++ b/hack.html.markdown
@@ -1,0 +1,307 @@
+---
+language: Hack
+contributors:
+    - ["Stephen Holdaway", "https://github.com/stecman"]
+filename: learnhack.hh
+---
+
+Hack is a superset of PHP that runs under a virtual machine called HHVM. Hack
+is almost completely interoperable with existing PHP code and adds a bunch of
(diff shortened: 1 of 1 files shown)


hack.html.markdown
```diff
@@ -0,0 +1,307 @@
...
+// Type hints must explicitly allow being nullable
+function identity(?string $stringOrNull) : ?string
+{
+    return $stringOrNull;
+}
+
+// Type hints can be specified on class properties
+class TypeHintedProperties
+{
+    public ?string $name;
```
levibostian: Is the `?` before string a typo?

stecman: No, this is Hack's syntax for denoting [nullable types](http://docs.hhvm.com/manual/en/hack.nullable.php). Any type hint can be prefixed with `?` to signal the type checker to allow a null value as well as the specified type.

levibostian:
Thanks!

Issue #957, opened 2015-02-06
The parser missed the lua-ru guide.


TheDmitry:
There is the lua-ru document in the ru-ru directory.
But it isn't shown on the LearnXinYMinutes homepage.

PR #966, opened 2015-02-18
[lua/ru] a probable solution #957

TheDmitry:
I'm not exactly familiar with LearnXinYMinutes-site parser. But if we replace the lowercase letter 'l' for a capital 'L', ru version of Lua is displayed on the main page. Please check this.


levibostian:
Thanks!

PR #967, opened 2015-02-18
[hack/en] Fix link to other learnxinyminutes article

stecman:
I had assumed a relative link to another markdown file in this project would work, but it renders as:

    http://learnxinyminutes.com/docs/hack/php.html.markdown

instead of the desired:

    http://learnxinyminutes.com/docs/php/

I've replaced this relative link with an absolute link to where the target page is located on the published site. This commit also fixes a typo in markdown syntax at the end of the document.

If there's an alternative way you'd prefer for this internal link to work, let me know.

### DIFF
--- a/hack.html.markdown
+++ b/hack.html.markdown
@@ -11,7 +11,7 @@ useful features from statically typed languages.
 
 
 Only Hack-specific features are covered here. Details about PHP's syntax are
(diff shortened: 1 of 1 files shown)


levibostian:
Thanks!

PR #968, opened 2015-02-18
[css/en] corrected definition of ~ selector, added |

awalGarg:
[attr~='foo'] will not select <el attr="foobar">, it will only select <el attr="bar foo baz">

Also added definition for [attr|='pre'] style selector

Refs: https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors, http://jsfiddle.net/awalGarg/72k3ptsk/


levibostian:
Thanks!

PR #971, opened 2015-02-21
Update README.markdown

gfvcastro:


levibostian:
I am not sure about this PR. I think it would be better if there was functionality built into the site that would download a PDF copy instead of advertising a tool you can use because there are tons of tools out there.

geoffliu:
Also, since all of our tutorials are single-paged, why can't you just print to PDF?

adambard:
It also seems like an endorsement with just the one option. I'd be more inclined to accept it if it also contained a link for pandoc or some other option.

PR #974, opened 2015-02-25
[Javascript/en] Add semicolon to maximine consistency.

qw3rtman:
Added a semicolon to conform with statement at top of guide:

// Because those cases can cause unexpected results, we'll keep on using
// semicolons in this guide.


levibostian:
Thanks!

PR #976, opened 2015-02-27
Update perl6.html.markdown

brxfork:
typo in with-optional() sub : semicolon / comma


vendethiel:
The code is correct: the argument is returned. The last expression of a function is implicitly returned. Should I clarify that point further in the tutorial?

brxfork:
Sorry for that. I was just confused because of the void-context of the call.
Perhaps it could be :
<code>
my $result-noarg  = with-optional; # returns (Any)
</code>
<code>
my $result-noarg2 = with-optional(); # returns (Any)
</code>
<code>
my $result1 = with-optional(1); # returns 1
</code>

Hmmm, not sure.

vendethiel:
well, it's void-context, but the comment says "returns ..."

PR #979, opened 2015-03-01
Fix Nim a bit

def-:
- The section on sequences was quite short, added a bit
- Fixed a typo
- Using proper Nim syntax highlighting. Is pygments used as highlighter?


vendethiel:
:+1: thanks !

PR #980, opened 2015-03-01
Perl6, minor change in comments

brxfork:
It seems "when" should be replace by "given" at the end of the comment line.


vendethiel:
Good catch! Thanks!

PR #981, opened 2015-03-01
Moved explanation of indexing into a list earlier

munrocape:
This way, it is clear what [1..] !! 999 means later on.


adambard:
Thanks!

PR #985, opened 2015-03-04
[git/en] Added the handy -a switch

awalGarg:
Added info about the `-a` switch which can come in handy at times.


geoffliu:
Thanks!

PR #986, opened 2015-03-04
Renamed Big-Oh to more prevalent name, Big-O.

Jakehp:
Everywhere I see it online references it as 'Big-O' or 'Big O'. Not sure why my books use Big-Oh. Don't want to cause any confusion with 'Big-Oh', so I renamed it all.


geoffliu:
Yup, good call!

Issue #996, opened 2015-03-10
Translations of the page headers


geoffliu:
... and footers too.

Even in the translated pages, the top of each page up to "Get the code" are still all in English. We should have translated version of those alongside, for user-friendliness, and probably SEO.

I'm thinking of either having one top-level config file that has the translated headers for every locale, or a special file in each of the locale specific directories. @adambard what do you think?

adambard:
That's definitely a great idea. It'll probably be a little while before I get to it, but I'll make sure it's next on my list.

PR #999, opened 2015-03-12
First attempt at getting a config file for page template strings

geoffliu:
@adambard This is first step toward #996, you can hook up to middleman whenever you get a chance. I used YAML since Ruby has decent support of it. I don't quite know how middleman does templating, so please comment. Especially on the line about contributors, since there are multiple substitutions, and they might not always be in the same order for different languages. Being able to pass in arguments by name would be nice.


geoffliu:
@adambard Gentle ping? :)

adambard:
Right! Fixed in adambard/learnxinyminutes-site@dbf5367fe4a0a7b2accfa0b0c30a26e24979b96b. Thanks for reminding me!

geoffliu:
Oh OK! How will people contribute to that though, being on a different (and less visible) repo?

adambard:
I made an issue at
https://github.com/adambard/learnxinyminutes-docs/issues/1053, and also
people can just post the translations or whatever in the thread.

PR #1001, opened 2015-03-13
Closes #1000

### LINKED ISSUE #1000: [c++/en] int main() and void main()
Though `void main()` is accepted by most compilers, it's not correct in both ANSI C and C++.
Therefore, it shouldn't be listed among the DIFFERENCES between them.

Arnie97:


vendethiel:
Thanks!

(and congrats on the #1000)

PR #1002, opened 2015-03-13
[racket/en] Add detailed explanation of eq?, eqv?, and equal?

keyanzhang:
This topic is confusing for a lot of people, so I think it’s worth to add it here. Thanks for reviewing!
CC: @elibarzilay @gustavoschmidt @cmpitg 

References:
1. http://docs.racket-lang.org/reference/booleans.html
2. http://docs.racket-lang.org/reference/eval-model.html
3. https://groups.google.com/forum/#!topic/plt-scheme/T1k49HMl450
4. http://stackoverflow.com/questions/16299246/what-is-the-difference-between-eq-eqv-equal-and-in-scheme
5. http://www.r6rs.org/final/html/r6rs/r6rs-Z-H-14.html

### DIFF
--- a/racket.html.markdown
+++ b/racket.html.markdown
@@ -7,6 +7,7 @@ contributors:
   - ["Eli Barzilay", "https://github.com/elibarzilay"]
   - ["Gustavo Schmidt", "https://github.com/gustavoschmidt"]
   - ["Duong H. Nguyen", "https://github.com/cmpitg"]
+  - ["Keyan Zhang", "https://github.com/keyanzhang"]
 ---
 
 Racket is a general purpose, multi-paradigm programming language in the Lisp/Scheme family.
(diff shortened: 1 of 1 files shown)


adambard:
I can see how that would be confusing. Thanks for the PR!

keyanzhang:
Thank you Adam!

elibarzilay:
As someone who actually teaches this stuff, the additional examples are not too great.  They demonstrate a bunch of things that happen to work now, but it's not clear whether they will continue working-- IOW, it's not clear what parts are language specification and what parts are due to the current implementation.  For example, right now Racket tries to avoid possibly confusing things like allocating a single `(3)` list when multiple expressions evaluate to it -- but can this change?  Another example: `(expt 2 100)` is currently a bignum, and therefore it's allocated -- but what happens when Racket run on some 128bit OS, or what if bignums that can be precomputed are cached?

Also, such changes are not theoretic.  For example, if you try this now:

    (define (foo x) (lambda (y) (+ y 3)))

then what would you get for `(eq? (foo 1) (foo 2))`?  The answers happens to depend on the compiler, and in this case it returns `#t` but that wasn't always the case.

The bottom line is that a very big disclaimer saying that this kind of pointer comparison can and often does expose implementation details would help a lot.  Specifically if it says that some of the listed examples might change with subtle changes to the compiler.  This applies to most uses of `eq?` that are done as an optimization (rather than uses that depend on object identity, like hash table caches).

OTOH, there are *valid* uses of `eq?` that are not mentioned at all.  One of the classic examples of this is using `eq?` on boxes to find out if two boxes are actually the same.

Finally, `eqv?` is mostly a confusing thing that was Scheme's attempt at providing a "mostly-useful and fast" equality, since many people would forget that `eq?` is not a good idea to compare numbers.    IMO (and in the opinion of many other schemers) it's mostly a failure.  (The thing is that `equal?` and `eq?` make sense as generic comparisons, `=` etc make sense as type-specific comparisons, that leaves `eqv?` in a similar state to Zaphod Beeblebrox: "it's just that guy, you know?.)

keyanzhang:
Thanks for reviewing, Eli! It's great to receive your feedback.

I agree—a huge disclaimer on `eq?`’s dependency on the compiler implementation will help (I did mean to emphasize it, but seems the way I present it is still not clear enough). I’ll edit it later. 

It’s interesting to know that `eqv?` is considered as a failure (I didn’t know about that!). Can I put it as “Use `eq?` when you know what you are doing; when in doubt, use `equal?`”? Also, are there any other valid/frequent uses of `eq?` worth mentioning, other than the box identity checking case?

In addition, could you point me to some documents/papers that list some Racket compiler implementation details on this topic (e.g., what stuffs are allocated and what are not), if there exists one? That may help in this case.

Again, thank you for reviewing! Nothing’s better than receiving feedback from a PLT member on this.

PR #1004, opened 2015-03-16
No-space array notation type[]

themiurgo:


adambard:
I would normally reject this on account of being a concern of coding style, but in this case I happen to agree. Hah!

PR #1006, opened 2015-03-16
Update haskell.html.markdown. Wrong explanation about '$' operator

nero-luci:
The right side of '$' is evaluated first, then the result is passed to the left side

### DIFF
--- a/haskell.html.markdown
+++ b/haskell.html.markdown
@@ -208,13 +208,13 @@ foo 5 -- 75
 -- to get rid of a lot of parentheses:
 
 -- before
-(even (fib 7)) -- true
+(even (fib 7)) -- false
 
 -- after
-even . fib $ 7 -- true
+even . fib $ 7 -- false
 
 -- equivalently
-even $ fib 7 -- true
+even $ fib 7 -- false
 
 ----------------------------------------------------
 -- 5. Type signatures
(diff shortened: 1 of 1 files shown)


geoffliu:
The previous wording wasn't great, but I don't think the new wording is any better. Both versions make the `$` sound like a special syntax, but in reality, it's just another operator. Since Haskell employs call-by-need evaluation, the new description isn't even accurate. The right side of `$` will only be evaluated *if* it is needed. 

I would rather have the explanation that `a $ b` is the same as `a b`, except that 1. `$` has really low precedence, and 2. `$` is right-associative. It has nothing to do with evaluation order.

vendethiel:
I agree.

nero-luci:
I agree that my explanation is not better, but is the current one correct?

geoffliu:
No it's not. It's kind of wrong in the same way. Haskell's evaluation order is based on value-dependency, not operator precedence. The `$` only makes a semantic difference, not a substantive one.

geoffliu:
No it's not. It's kind of wrong in the same way. Haskell's evaluation order is based on value-dependency, not operator precedence. The `$` only makes a semantic difference, not a substantive one.

nero-luci:
@geoffliu is the latest explanation correct? 
Thank you!

geoffliu:
Hey @nero-luci, thanks for putting up with me! :)

This is much better, although I still wouldn't use the word "evaluate" in there. It's really awkward, I know; I have a hard time thinking of a good wording too. I just checked on LYAHFGG, he explained it as "the expression on its right is applied as the parameter to the function on its left." Maybe we could write something similar to that?

geoffliu:
Hey thanks, both for the work and for putting up with me. :)

nero-luci:
Hey, thank you for useful feedback and patience. It was pleaseful :)

PR #1007, opened 2015-03-16
[Haskell/en] Make the two fib functions consistent

geoffliu:
Noticed when reviewing #1006, the two fib definitions are off by one.


vendethiel:
good catch

Issue #717, opened 2014-08-09
Functions recommended


Argzero:
Based off of ruby based implementations:

- Dictionaries / Hashes should have a has_key? (or for julia, hasKey) and has_value? (hasValue) function
- Arrays ought to have an include? function or equivalent

bharrisau:
Is this closed by #1011?

vendethiel:
I'd think so, yes.

PR #1011, opened 2015-03-19
[ruby] Array and Hash specific functions

iamrahil:
In reference to https://github.com/adambard/learnxinyminutes-docs/issues/717
Adding has_key? and has_value? methods for Hash
Adding include? method for Array

Note: I have used paranthesis while giving examples of these functions because the fact that they are optional is stated far below. It might create confusion if someone is reading from top to bottom.


levibostian:
Thanks!

PR #1013, opened 2015-03-23
fixes for haxe

nadako:


haxe.html.markdown
```diff
@@ -11,7 +12,7 @@ Haxe author).  Note that this guide is for Haxe version 3.  Some of the guide
 may be applicable to older versions, but it is recommended to use other
 references.
 
-```csharp
+```haxe
```
levibostian: Sorry, but haxe [is not supported](https://github.com/adambard/rouge/tree/master/lib/rouge/lexers) in the website. Will have to use csharp as it is the best alternative.

haxe.html.markdown
```diff
@@ -34,16 +35,20 @@ references.
 /*
    This is your first actual haxe code coming up, it's declaring an empty
    package.  A package isn't necessary, but it's useful if you want to create a
-   namespace for your code (e.g. org.module.ClassName).
+   namespace for your code (e.g. org.yourapp.ClassName).
+
+   Omitting package declaration is the same as declaring empty package.
```
levibostian: `declaring empty package` => `declaring an empty package`

haxe.html.markdown
```diff
@@ -383,11 +395,7 @@ class LearnHaxe3{
         */
 
         // if statements
-        var k = if (true){
-            10;
-        } else {
-            20;
-        }
+        var k = if (true) 10 else 20;
```
levibostian: Is using one line for conditionals in the style guidelines?

haxe.html.markdown
```diff
@@ -628,6 +636,7 @@ enum ComplexEnum{
     ComplexEnumEnum(c:ComplexEnum);
 }
 // Note: The enum above can include *other* enums as well, including itself!
+// Note: This is what called *Algebraic data type* in some other languages.
```
levibostian: `what` => `whats`

levibostian:
Good to go after checking over those quick notes I made. Thanks!

haxe.html.markdown
vendethiel: (`what's`, even)

haxe.html.markdown
nadako: It's somehow not obvious that one can use more lightweight syntax without curly braces. I think this guide also need to mention that even functions don't require `{}`, but maybe later.

nadako:
Done.

levibostian:
Thanks!

PR #1016, opened 2015-03-24
[python3.html.mardown] Added a short statement about magic methods

nero-luci:
Terminology related to Python special functions


geoffliu:
Agreed. Thanks!

PR #1022, opened 2015-03-30
c.html: fix #1021 (bitwise negation and shifting into the sign bit)

### LINKED ISSUE #1021: C: negation and shift
> ~0x0F; // => 0xF0 (bitwise negation, "1's complement")

This looks like byte negation, but 0x0f is of type int. A typical result could be 0xfffffff0.

>// - shifting into the sign bit of a signed integer (int a = 1 << 32)

Should be (int a = 1 << 31), as 31'st bit is the sign bit of a 32-bit wide int.

jcmvbkbc:
0x0f is of type int, for 32-bit int the result is 0xfffffff0.
31'st bit is the sign bit of a 32-bit wide int.

### DIFF
--- a/c.html.markdown
+++ b/c.html.markdown
@@ -234,7 +234,7 @@ int main() {
   // same with j-- and --j
 
   // Bitwise operators!
-  ~0x0F; // => 0xF0 (bitwise negation, "1's complement")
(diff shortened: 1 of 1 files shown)


geoffliu:
Nice catch! Thanks!

PR #1025, opened 2015-03-30
[erlang/en] Fix syntax error in erlang if-expression

nomatteus:
As required by Erlang:
http://erlang.org/doc/reference_manual/expressions.html#id78310


vendethiel:
that's incorrect. The last one doesn't need a semicolon, as the `end` will close the if anyways.

vendethiel:
It seems I read that diff the other way around. my bad :)! thanks for the PR

PR #1030, opened 2015-04-01
Assembly language

Th3G33k:


assembly.html.markdown
```diff
@@ -0,0 +1,218 @@
+---
+language: assembly
+filename: learn.asm
+contributors:
+    - ["", ""]
```
vendethiel: ?

assembly.html.markdown
```diff
@@ -0,0 +1,218 @@
...
+---
+
+Assembly language is a low-level programming language for a computer
```
vendethiel: which asm is that? :P

Th3G33k:
x86 asm, intel syntax (nasm)

assembly.html.markdown
Th3G33k: x86 asm, intel syntax

assembly.html.markdown
```diff
@@ -0,0 +1,218 @@
...
+
+```asm
+; Comments
```
geoffliu: A little more verbose here. As a novice, I'm not sure if this is special syntax, whether "Comments" has special meaning like "section" does. Maybe "Anything from semicolon until the end are comments"?

assembly.html.markdown
```diff
@@ -0,0 +1,218 @@
...
+
+section .data
+; it contains data, constants
```
geoffliu: `s/it/It/`. Also could you give a few examples of C code that would generate this section? (String literals, etc.)

assembly.html.markdown
```diff
@@ -0,0 +1,218 @@
...
+doubleword_ dw 0 ; 4 bytes
+; string zero terminated
+msg db "hello, world", 0
+; array of short int
+array dw 0, 1, 2, 3, 4
+
+section .text
+; it contains all the instructions to be executed
+
+; where the program execution begins
```
geoffliu: It might be helpful to mention how `_start` relates to `main` in C.

assembly.html.markdown
geoffliu: Also, are any parts of this guide 32-bit specific?

assembly.html.markdown
```diff
@@ -0,0 +1,218 @@
...
+```asm
+; Comments
+
+section .data
+; it contains data, constants
+
+; (variable-name) define-directive initial-value
+byte_ db 0 ; 1 byte
+word_ dw 0 ; 2 bytes
+doubleword_ dw 0 ; 4 bytes
```
geoffliu: `s/dw/dd/`

In general, this section is pretty confusing. Could you add a little more explanation about what each part of the line means, what `db`, `dw`, `dd` mean, and that it's OK to specify multiple values, and they all get turned into an array. Also why `()` around variable name.

Side note, I don't see a `.rodata` section, but I think it's worth mentioning.

assembly.html.markdown
```diff
@@ -0,0 +1,218 @@
+---
+language: assembly
+filename: learn.asm
```
geoffliu: Isn't `.s` the usual extension for assembly on Linux?

assembly.html.markdown
```diff
@@ -0,0 +1,218 @@
...
+movsb ; movsb, movsw, movsd, [edi] = [esi]
+lodsb ; eax = [esi]
+scasb ; cmp eax, [edi]
+cmpsb ; cmp [esi], [edi]
+stosb ; [edi] = eax
+
+; CDECL calling convention
+
+; _fct (2, 3)
+push 3 ; pass arguments right to left
```
geoffliu: Would be nice to mention the benefits of going right to left, var-arg functions, for example.

assembly.html.markdown
```diff
@@ -0,0 +1,218 @@
...
+; 1 exit(int error_code)
+; 3 read(uint fd, char *buf, int count)
+; 4 write(uint fd, char *buf, int count)
+; 5 open(char *filename, int flags, int mode)
+; 6 close(uint fd)
+; 8 creat(char *pathname, int mode)
+; 13 time(int *tloc)
+; 45 sys_brk(int brk) allocate memory
+
+mov eax, 4 ; write
```
geoffliu: Maybe explain the calling convention here. It's quite different from the C-convention, after all.

assembly.html.markdown
```diff
@@ -0,0 +1,218 @@
...
+mov ebx, 1 ; stdout
+mov ecx, msg
+mov edx, 13
+int 80h
```
geoffliu: Explain what `int 80` actually means. Most programmers aren't even aware of the user-space-kernel-space divide, let alone what interrupts mean.

assembly.html.markdown
```diff
@@ -0,0 +1,218 @@
...
+neg eax ; eax = -eax
+
+; LOGICAL
+
+and eax, ebx ; eax &= ebx
+or eax, [doubleword_] ; eax |= [doubleword_]
+xor [doubleword_], eax ; [doubleword_] ^= eax
+not eax ; eax = !eax
+
+mov al, 0b11001100
```
geoffliu: Add a "SHIFTS" section header here.

geoffliu:
Hey @Th3G33k, thanks for the work! With assembly, I think there needs to be a lot more words than code, since a deep understanding of the hardware, kernel, and their interactions is required, hence all the comments. Please bear with us; we want this guide to be beginner friendly. :)

Th3G33k:
Thank you for the review.
I will change the code and I will make a new commit.

assembly.html.markdown
geoffliu: Actually, not string literals. This would be static variables, right?

assembly.html.markdown
adambard: To clarify, this is where your name (real or whatever) and website should go, e.g. `["Th3G33k", "https://github.com/Th3G33k"]`

PR #1031, opened 2015-04-01
Update perl6.html.markdown // when and smartmatch

brxfork:
($_.chars > 50) ~~ True : this is always True.


vendethiel:
Good catch!

PR #1036, opened 2015-04-11
[c++/en] Fix spelling

Arnie97:


geoffliu:
Yup, that does seem wrong. Thanks for the fix!

PR #1044, opened 2015-04-21
Add default case to switch statement in Go

thisiskevinchoi:


levibostian:
I almost added a note saying to please add another comment to the default case explaining what the default case did but I am assuming that people using these docs are already programmers so they should know what a default case is. 

Thanks!

PR #1045, opened 2015-04-21
fix the while macro for Common Lisp

bcc32:
Earlier version was incorrect, as it only executed `body` once and then terminated, like an if statement instead.


levibostian:
Thanks! Great catch.

PR #1047, opened 2015-04-22
Fixed grammar, added explanation of function signature and arguments

SamuelMarks:


levibostian:
Thanks!

go.html.markdown
```diff
@@ -177,6 +181,10 @@ func learnFlowControl() {
 	case 1:
 	case 42:
 		// Cases don't "fall through".
+		/*
```
vendethiel: It feels weird to have one comment inline and the other one in a block comment

PR #1050, opened 2015-04-22
[python3/en] Adding strings without + added.

srirams6:
Added for Python3 docs as well.


srirams6:
@levibostian Added the same for python3 as well!

geoffliu:
Thanks!

PR #1051, opened 2015-04-22
[python/en] [python3/en] Added range, correct checking for True/False.

srirams6:
Added `range(start, stop)` for Python2.
Added examples for comparisons such as `0 is False` which is the preferred way to compare objects to True/False in Python.

### DIFF
--- a/python.html.markdown
+++ b/python.html.markdown
@@ -133,8 +133,15 @@ None is None  # => True
 
 # None, 0, and empty strings/lists all evaluate to False.
 # All other values are True
-bool(0)  # => False
+bool(0)   # => False
 bool("")  # => False
+# Don't use the equality symbol to check if an object is True or False
+# Use "is" instead. This checks for equality of object identity.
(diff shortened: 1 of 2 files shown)


python.html.markdown
```diff
@@ -133,8 +133,15 @@ None is None  # => True
...
 # All other values are True
-bool(0)  # => False
+bool(0)   # => False
 bool("")  # => False
+# Don't use the equality symbol to check if an object is True or False
+# Use "is" instead. This checks for equality of object identity.
+0 is True   # => False
+0 is False  # => True
+1 is True  # => True
+[] is False # => True
```
geoffliu: First of all, you shouldn't use equality symbols to check truthiness at all. Just use the expression itself. Instead of `if x == True`, just say `if x`.

Secondly, `is` is checking for object equality. `[] is False` cannot possibly be true. I suspect the other statements are wrong too.

PR #1054, opened 2015-04-23
[python/en] Range function arguments.

srirams6:
Range function with start and stop.

### DIFF
--- a/python.html.markdown
+++ b/python.html.markdown
@@ -347,6 +347,18 @@ for i in range(4):
     print(i)
 
 """
+"range(lower, upper)" returns a list of numbers
+from the lower number to the upper number
+prints:
+    4
+    5
+    6
+    7
+"""
+for i in range(4, 8):
+    print(i)
+
+"""
 While loops go until a condition is no longer met.
 prints:
     0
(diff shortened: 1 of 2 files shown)


levibostian:
Could you add the original Python doc changes to this PR as well? Thanks!

levibostian:
Sorry it is a pain we have 2 Python docs. Thanks for the contributions!n

srirams6:
@levibostian  Done :)

python.html.markdown
```diff
@@ -347,6 +347,18 @@ for i in range(4):
...
+"range(lower, upper)" returns a list of numbers
+from the lower number to the upper number
+prints:
+    4
+    5
+    6
+    7
+"""
+for i in range(4, 8):
+    print(i)
```
geoffliu: Ah but this is Python 2. `print i` please. :)

srirams6: The whole doc uses the print_function from future. This isn't a good way to teach/learn python2 in my opinion, but I was just sticking to the format of the existing doc.

geoffliu: Good call, I hadn't realized. Thanks for the PR, and I see you've opened an issue about print statements.

PR #1055, opened 2015-04-23
initial version for latex

icymist:
Initial version for LaTeX.

First time requesting for a pull. Let me know if something is not right.

### DIFF
--- a/latex.html.markdown
+++ b/latex.html.markdown
@@ -1,0 +1,133 @@
+---
+language: latex
+contributors:
+    - ["Chaitanya Krishna Ande", "http://icymist.github.io"]
+filename: learn-latex.tex
+---
+
+LaTeX is known to create aesthetically pleasing documents without you worrying
+about the formatting. It is also great if one wants to create documents
+containing a lot of mathematics. Getting a good document is very easy, but
+getting it to behave exactly the way you want can be a bit hairy.
+
+
+```latex
+% All comment lines start with %
+% There are no multi-line comments
+
+% LaTeX is NOT a ``What You See Is What You Get'' word processing software like
+% MS Word, or OpenOffice Writer
+
+% Getting to the final document using LaTeX consists of the following steps:
+% 1. Write the document in plain text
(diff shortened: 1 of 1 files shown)


latex.html.markdown
```diff
@@ -0,0 +1,133 @@
...
+about the formatting. It is also great if one wants to create documents
+containing a lot of mathematics. Getting a good document is very easy, but
+getting it to behave exactly the way you want can be a bit hairy.
+
+
+```latex
+% All comment lines start with %
+% There are no multi-line comments
+
+% LaTeX is NOT a ``What You See Is What You Get'' word processing software like
```
levibostian: You are using a combination of `` and '' here. Do 'What You See Is What You Get'

latex.html.markdown
```diff
@@ -0,0 +1,133 @@
...
+% MS Word, or OpenOffice Writer
```
levibostian: If Latex is not this, what is it?

latex.html.markdown
```diff
@@ -0,0 +1,133 @@
...
+% You write all your formatting information in plain text in Step 1.
+% The compilation part in Step 2 takes care of producing the document in the
+% format you defined in Step 1.
+
+% For Step 1, it is best if you get a good text editor
+% On Windows, probably Notepad++
+% For Step 2, you will need to get a TeX distribution
+% Windows: MikTeX
+% MacOS: MacTeX
+% Linux: Should be available from your package manager
```
levibostian: What is the package called?

levibostian:
The document flow is kind of confusing. This seems more like a blog post tutorial telling you how to get LaTeX up and running. LearnXinYminutes is usually used for a quick learning tool or reference. 

For example, the part where you talk about the document{begin} and document{end}, there is a lot of repeating there. Just state document{begin} once and then state what that means. Make it flow a little better. 

Make sense?

geoffliu:
Building on what @levibostian is saying, I think it would be nice to break this doc into sections using the `\section`s, rather than with comments, as we do with other docs.

PR #1059, opened 2015-04-24
[python/en] Moved to default python2 print statements.

### LINKED ISSUE #1058: [python/en] correct usage of print
The entire doc uses `print("test")` instead of `print "test"`. Even though `print_function` is imported from `__future__`, this isn't the default print statement in python 2. It would confuse people that are new to the language. I propose to change all the print statements to the default python 2 print. 

If this is issue is in order, I shall send a PR.

srirams6:
Fixes #1058

### DIFF
--- a/python.html.markdown
+++ b/python.html.markdown
@@ -141,14 +141,8 @@ bool("")  # => False
 ## 2. Variables and Collections
 ####################################################
 
-# Python has a print statement, in all 2.x versions but removed from 3.
+# Python has a print statement
 print "I'm Python. Nice to meet you!"
-# Python also has a print function, available in versions 2.7 and 3...
-# but for 2.7 you need to add the import (uncommented):
-from __future__ import print_function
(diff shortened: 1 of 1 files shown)


python.html.markdown
```diff
@@ -145,8 +145,10 @@ bool("")  # => False
 print "I'm Python. Nice to meet you!"
 # Python also has a print function, available in versions 2.7 and 3...
 # but for 2.7 you need to add the import (uncommented):
-# from __future__ import print_function
+from __future__ import print_function
```
geoffliu: I would say let's just get rid of the `__future__` discussion entirely. There are lots of Python3 features that can be enabled from a `__future__` import, kind of odd that we're singling out printing.

srirams6: :+1: Done.

geoffliu:
Nice, thanks!

PR #1061, opened 2015-04-24
[python/en] Added note about generator.

srirams6:


python.html.markdown
```diff
@@ -568,7 +568,7 @@ dir(math)
 ## 7. Advanced
 ####################################################
 
-# Generators help you make lazy code
+# Generators help you make efficient code
```
geoffliu: "Lazy" was the precise description for generators. It explains that the values are lazily computed. "Efficient," on the other hand, is much more general, and therefore less descriptive.

python.html.markdown
```diff
@@ -577,6 +577,8 @@ def double_numbers(iterable):
 # Instead of generating and returning all values at once it creates one in each
 # iteration.  This means values bigger than 15 wont be processed in
 # double_numbers.
+# Also, it means that the entire list won't be stored in memory.
```
geoffliu: Generators are not lists. Frankly, I think the lines before and after already explains the concept pretty well.

PR #1065, opened 2015-04-26
[standard-ml] Standard ML has references!

j-baker:
Intro misrepresents SML by suggesting it has no references. Fixed buggy wording of introduction and added sample usages of references, while and seq.


vendethiel:
Looks good, thanks!

PR #1069, opened 2015-04-26
[bash/en] Add another common bash variable assignment mistake

deryni:


bash.html.markdown
```diff
@@ -36,7 +37,14 @@ VARIABLE="Some string"
 # But not like this:
 VARIABLE = "Some string"
 # Bash will decide that VARIABLE is a command it must execute and give an error
-# because it couldn't be found.
+# because it can't be found.
+
+# Or like this:
+Variable= 'Some string'
```
levibostian: Could you please all caps the 'Variable' name here to match the example? The PR you made with all caps can then discuss if we should keep it all caps or not but for this PR, lets do all caps please to keep the examples working. 

Thanks!

deryni:
Done.

Sorry, if I'd been making these changes independently I'd have followed case here correctly in the first place but I added this locally while making the ALL_CAPS changes (among other changes; PR's for those will be coming later today probably).

levibostian:
Awesome! Thank you and thanks for all of your contributions so far. We really appreciate it!

PR #1070, opened 2015-04-26
[bash/en] Don't use ALL_CAPS variable names.

deryni:
ALL_CAPS variable names are traditionally "reserved" for use by the
shell/system.

(People often try to use PATH for things locally and then wonder why their
commands all stop working for example.)

### DIFF
--- a/bash.html.markdown
+++ b/bash.html.markdown
@@ -32,40 +32,40 @@ echo Hello world!
 echo 'This is the first line'; echo 'This is the second line'
 
 # Declaring a variable looks like this:
-VARIABLE="Some string"
+Variable="Some string"
 
 # But not like this:
-VARIABLE = "Some string"
-# Bash will decide that VARIABLE is a command it must execute and give an error
+Variable = "Some string"
+# Bash will decide that Variable is a command it must execute and give an error
 # because it can't be found.
 
 # Or like this:
-VARIABLE= 'Some string'
+Variable= 'Some string'
 # Bash will decide that 'Some string' is a command it must execute and give an
-# error because it can't be found. (In this case the 'VARIABLE=' part is seen
+# error because it can't be found. (In this case the 'Variable=' part is seen
 # as a variable assignment valid only for the scope of the 'Some string'
 # command.)
 
 # Using the variable:
-echo $VARIABLE
-echo "$VARIABLE"
-echo '$VARIABLE'
+echo $Variable
+echo "$Variable"
+echo '$Variable'
 # When you use the variable itself — assign it, export it, or else — you write
 # its name without $. If you want to use variable's value, you should use $.
 # Note that ' (single quote) won't expand the variables!
 
 # String substitution in variables
-echo ${VARIABLE/Some/A}
+echo ${Variable/Some/A}
 # This will substitute the first occurance of "Some" with "A"
 
 # Substring from a variable
-LENGTH=7
-echo ${VARIABLE:0:LENGTH}
+Length=7
+echo ${Variable:0:Length}
 # This will return only the first 7 characters of the value
 
 # Default value for variable
-echo ${FOO:-"DefaultValueIfFOOIsMissingOrEmpty"}
-# This works for null (FOO=) and empty string (FOO=""); zero (FOO=0) returns 0.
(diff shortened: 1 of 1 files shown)


levibostian:
Interesting. I have never seen variables not be all caps for bash variables. I should never say never but I do see all caps pretty much all the time online for scripts I see. I know it doesn't really matter, I am just thinking about the bash style guidelines.

deryni:
Then I'd claim you aren't looking in very good places. =)

But I wouldn't be overly surprised by that fact actually. There is a *huge* amount of poor-to-bad bash/shell scripting code out there.

Even the generally not so correct/safe/good TLDP Advanced Bash-Scripting Guide doesn't use ALL_CAPS variables (and they do use all sorts of other bad/unsafe practices).

The [POSIX spec](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html#tag_08_01) says:

> Environment variable names used by the utilities in the Shell and Utilities volume of POSIX.1-2008 consist solely of uppercase letters, digits, and the <underscore> ( '_' ) from the characters defined in Portable Character Set and do not begin with a digit.

So using those means you *might* conflict and not using those means you *can't* conflict. The actual set of ALL_CAPS variables that you need to worry about is fairly small but people do run into them (you can search on [stackoverflow](https://stackoverflow.com/questions/tagged/bash) to see this happen with `PATH` with some, not frequent, but still regular regularity).

levibostian:
I completely understand that. What tutorials, books, operating systems have bash code that is not all caps? I am still afraid to merge as I still don't see it is convention. I don't want the site to be the odd one out there on the Internet.

deryni:
I would be surprised if you could find one example of ALL_CAPS variables on the bashfaq/bashguide pages. Nearly *all* of the code I see on the bash tag on SO is non-ALL_CAPS (and this is including the code coming from users asking the questions). Code that does contain ALL_CAPS variables is routinely "corrected" in an answer not to do that (and the OP is given the same advice to avoid it).

Many of the examples in the [bash reference](http://www.gnu.org/software/bash/manual/bashref.html) avoid ALL_CAPS variables (though some use them for no reason I can see).

There is an additional convention (that I'm not a huge fan of but does exist) which suggests that intentionally global shell variables (those that get exported/etc.) should also be ALL_CAPS (to match the system/shell variables that are exported by default. I understand the argument and even mostly agree with it. And done correctly it is likely a good convention. But in cases designed to teach people unfamiliar with the shell (like SO and here and my own scripts that may be read by people less familiar than myself) I try to avoid using them.

If I could I'd like to turn the question around on you and ask you what sites you use that regularly use ALL_CAPS variables in their scripts?

levibostian:
I am a fan of http://www.cyberciti.biz/ for bash commands. Like http://www.cyberciti.biz/faq/bash-loop-over-file/ to loop over files in a directory. Besides that, I use SO to get quick references. Time to time I use http://tldp.org/. 

I do understand your point and I do see the problem for a beginner if they were to collide a local variable with a system variable and wonder what is going on. 

Instead of changing all the variables in the doc to not all caps, could we leave them alone and create some examples showing variables colliding with each other (like the PATH variable) and why they collide. Then show how to prevent colliding and how variables are case sensitive in bash?

deryni:
The majority of variables on that looping page are lower case. Only a very few are ALL_CAPS (and they are explicitly not the "normal" variables).

That page also while **specifically** telling you to be sure that you always use `"$f"` (and not `$f`) gets it wrong by using `$@` instead of `"$@"` in the loop.

It also uses the very bad practice of putting multiple "words" into a string and then expecting shell word-splitting to give them the words back (the `$FILES` example at the top) that is not suitable for use with any files/words that contain spaces or shell glob/metacharacters.

Example of collision and case-sensitivity are both likely good ideas but aren't really related to the issue here which is that using ALL_CAPS variables for "normal" variables is bad practice.

levibostian:
I do enjoy those sites for reference. 

I just did some research (which I probably should have done days ago) on conventions for lower case vs upper case and the community agrees that lower case is preferred. Lower case it is, lets merge this thing. 

Now only issue is we have merge conflicts...could you merge master and update please?

deryni:
Bah, lost track of this for a bit. Rebased off of master.

levibostian:
Thanks!

PR #1076, opened 2015-04-30
Fix Python typo.

rinoc:
Changed "Revert" to "Reverse".


geoffliu:
I've been meaning to get to this one for a while now, thanks! One small request: the current wording is really misleading. The slicing actually returns a copy of the original list, but the current wording reads as if the list is reversed in place. Could you clarify a bit, while we're at it?

rinoc:
Sure, how does this sound: "Returns a list with items in reverse order"?

geoffliu:
"a copy of the list"? :)

rinoc:
Alright, changed to: "Reverse a copy of the list". :smile:

geoffliu:
Ha fine. No promise I won't change it again though. :P

PR #1084, opened 2015-05-03
[julia/en] Update for Julia 0.3 and fix a few typos

garrison:


julia.html.markdown
```diff
@@ -190,7 +190,7 @@ end
 # inside the julia folder to find these files.
 
 # You can initialize arrays from ranges
-a = [1:5] # => 5-element Int64 Array: [1,2,3,4,5]
+a = [1:5;] # => 5-element Int64 Array: [1,2,3,4,5]
```
geoffliu: Hmm, isn't the semicolon only necessary when you want a new row? Or is this part of the "0.3 update"?

garrison: This is fine in 0.3.  This specific change is to future-proof it for 0.4, actually, where such concatenation is deprecated.

garrison: In julia 0.4:

    julia> [1:5]
    WARNING: [a] concatenation is deprecated; use [a;] instead

geoffliu:
Great, merging this now. Thanks!

PR #1089, opened 2015-05-04
[C++/en] Templates and gotchas

geoffliu:


c++.html.markdown
```diff
@@ -433,6 +433,82 @@ int main () {
...
+
+// Template parameters don't have to be classes:
+template<int Y>
+void printMessage() {
+  cout << "Learn C++ in " << Y << " minutes!" << endl;
+}
+
+// And you can explicitly specialize templates for more efficient code (most
+// real-world uses of specialization are not as trivial as this):
+template<>
```
vendethiel: Please add a comment for that necessary empty template

c++.html.markdown
```diff
@@ -585,6 +661,33 @@ void doSomethingWithAFile(const std::string& filename)
...
+  virtual void bar();
+};
+class FooSub : public Foo {
+  virtual void bar();  // overrides Foo::bar!
+};
+
+// 0, false, NULL are all the same thing!
+bool* pt = new bool;
+*pt = 0;  // Sets the value points by 'pt' to false.
+pt = 0;  // Sets 'pt' to the null pointer. Yes both lines compile without warning.
```
vendethiel: Please add a note about nullptr (if there's none in the tutorial already)

c++.html.markdown
```diff
@@ -585,6 +661,33 @@ void doSomethingWithAFile(const std::string& filename)
...
+
+// '=' != '='
+Foo f1 = f2;  // Calls Foo::Foo(const Foo&) or some variant copy constructor.
+
+Foo f1;
+f1 = f2;  // Calls Foo::operator=(Foo&) or variant.
```
vendethiel: Maybe you could use `FooSub` to demonstrate object slicing?

vendethiel:
:+1:

geoffliu:
Done and done. @vendethiel what do you think?

c++.html.markdown
```diff
@@ -585,6 +663,54 @@ void doSomethingWithAFile(const std::string& filename)
...
+*pt = 0;  // Sets the value points by 'pt' to false.
+pt = 0;  // Sets 'pt' to the null pointer. Both lines compile without warnings.
+
+// nullptr is supposed to fix some of that issue:
+int* pt2 = new int;
+*pt2 = nullptr;  // Doesn't compile
+pt2 = nullptr;  // Sets pt2 to null.
+
+// But somehow 'bool' type is an exception.
+*pt = nullptr;  // This still compiles, even though '*pt' is a bool!
```
vendethiel: I guess it's safe to say it's an exception so that `if (nullptr)` works?

c++.html.markdown
```diff
@@ -585,6 +663,54 @@ void doSomethingWithAFile(const std::string& filename)
...
+// '=' != '=' != '='!
+// Calls Foo::Foo(const Foo&) or some variant copy constructor.
+Foo f2;
+Foo f1 = f2;
+
+// Calls Foo::Foo(const Foo&) or variant, but only copies the 'Foo' part of
+// 'fooSub'. Any extra members of 'fooSub' are discarded. This lovely behavior
+// is called "object slicing."
+FooSub fooSub;
+Foo f1 = fooSub;
```
vendethiel: Maybe we could find a link explaining why that might be an issue? (stackoverflow or so)

c++.html.markdown
geoffliu: Yes that is the reason behind it, but still surprising. :)

c++.html.markdown
geoffliu: Eh I don't know. I left the name in there so people can search for it if they want to find out more. We're not meant to be a textbook that explains everything.

c++.html.markdown
vendethiel: Well, I think adding the explanation might be worth it. LGTM otherwise

c++.html.markdown
vendethiel: then maybe changing "lovely behavior" with something that says "here be snakes" would be good?

geoffliu:
@vendethiel Updated the comments. I guess you're right, sarcasm with "lovely" might not always come across. :)

vendethiel:
Amazing!

PR #1096, opened 2015-05-13
[Perl 6] tweak. Parrot desupported on Rakudo.

dwarring:
Also mention MoarVM before JVM.

### DIFF
--- a/perl6.html.markdown
+++ b/perl6.html.markdown
@@ -10,8 +10,7 @@ contributors:
 Perl 6 is a highly capable, feature-rich programming language made for the
 upcoming hundred years.
 
-Perl 6 runs on [the Parrot VM](http://parrot.org/), the JVM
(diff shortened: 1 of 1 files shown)


levibostian:
On [their website](http://www.parrot.org/) it states that Perl 6 is supported. Where do you find it being desupported?

dwarring:
Hi @levibostian,
From the release notes from latest rakudo star release: https://github.com/rakudo/rakudo/blob/nom/docs/announce/2015.03.md

"this is the first release of Rakudo that does not support Parrot [^2] as a backend. Whether Parrot support can be restored at some point in the future, really depends on people with tuits making the necessary changes. In that light, I would like to emphasize the last paragraph of said blog post:"

vendethiel:
Thanks!

PR #1101, opened 2015-05-18
[erlang/en] Various (mostly cosmetic) improvements

Jubobs:
* Add missing square brackets in example of pattern matching on lists
* Keep lines under 80 characters
* Hiphenate complex modifiers (e.g. "pattern-matching operation")
* Consistently proper case the language name
* Improve punctuation in comments
* Properly format Markdown inline code
* Tense changes (where appropriate)


levibostian:
Thanks!

PR #1103, opened 2015-05-19
[haxe/en] Correct the condition variable

sm4rk0:
...for "do statement always runs at least once" section.

Note: This is my first pull request on GitHub, so please be gentle if I made a mistake (:

### DIFF
--- a/haxe.html.markdown
+++ b/haxe.html.markdown
@@ -323,7 +323,7 @@ class LearnHaxe3{
         var  l = 0;
         do{
             trace("do statement always runs at least once");
-        } while (i > 0);
+        } while (l > 0);
 
         // for loop
         /*


levibostian:
Thanks!  That is a good catch so thank you for finding that error. 

Welcome to LearnXinYMinutes and also welcome to GitHub! Glad to have you here. You did great. 

As you get more involved with git and GitHub one thing you will want to improve on if your commit messages. So here, you made one commit and the message is "Update haxe.html.markdown" which works but one that would be better would be "Fix do while loop in haxe doc by renaming to correct variable" to be more descriptive of what you actually changed.

sm4rk0:
You're right. That message was auto-generated by GitHub's web interface. Thanks for feedback!

PR #1106, opened 2015-05-20
[erlang/en] Fix #970

### LINKED ISSUE #970: [erlang] wrong example of guards:
this example is making wrong expectations: 
```erlang
% A `guard sequence` is either a single guard or a series of guards, separated
%by semicolons (`;`). The guard sequence `G1; G2; ...; Gn` is true if at least
% one of the guards `G1, G2, ...` evaluates to true.
is_pet(A) when is_dog(A); is_cat(A) -> true;
is_pet(A) -> false.
```
there is no way to use as guard custom defined functions:

http://erlang.org/doc/reference_manual/expressions.html

```
...
8.25  Guard Sequences

A guard sequence is a sequence of guards, separated by semicolon (;). The guard sequence is true if at least one of the guards is true. (The remaining guards, if any, will not be evaluated.)
Guard1;...;GuardK

A guard is a sequence of guard expressions, separated by comma (,). The guard is true if all guard expressions evaluate to true.
GuardExpr1,...,GuardExprN

The set of valid guard expressions (sometimes called guard tests) is a subset of the set of valid Erlang expressions. The reason for restricting the set of valid expressions is that evaluation of a guard expression must be guaranteed to be free of side effects. Valid guard expressions are:

the atom true,
other constants (terms and bound variables), all regarded as false,
calls to the BIFs specified below,
term comparisons,
arithmetic expressions,
boolean expressions, and
short-circuit expressions (andalso/orelse).

is_atom/1
is_binary/1
is_bitstring/1
is_boolean/1
is_float/1
is_function/1
is_function/2
is_integer/1
is_list/1
is_map/1
is_number/1
is_pid/1
is_port/1
is_record/2
is_record/3
is_reference/1
is_tuple/1

Table 8.4:   Type Test BIFs.

Note that most type test BIFs have older equivalents, without the is_ prefix. These old BIFs are retained for backwards compatibility only and should not be used in new code. They are also only allowed at top level. For example, they are not allowed in boolean expressions in guards.

abs(Number)
bit_size(Bitstring)
byte_size(Bitstring)
element(N, Tuple)
float(Term)
hd(List)
length(List)
map_size(Map)
node()
node(Pid|Ref|Port)
round(Number)
self()
size(Tuple|Bitstring)
tl(List)
trunc(Number)
tuple_size(Tuple)

Table 8.5:   Other BIFs Allowed in Guard Expressions.

If an arithmetic expression, a boolean expression, a short-circuit expression, or a call to a guard BIF fails (because of invalid arguments), the entire guard fails. If the guard was part of a guard sequence, the next guard in the sequence (that is, the guard following the next semicolon) will be evaluated.
...
```

Jubobs:


erlang.html.markdown
```diff
@@ -167,8 +167,12 @@ is_dog(A) -> false.
...
 % by semicolons (`;`). The guard sequence `G1; G2; ...; Gn` is true if at
 % least one of the guards `G1`, `G2`, ..., `Gn` evaluates to `true`.
-is_pet(A) when is_dog(A); is_cat(A) -> true;
-is_pet(A) -> false.
+is_pet(A) when is_atom(A), (A =:= dog) or (A =:= cat) -> true;
+is_pet(A)                                             -> false.
+
+% Warning: not all valid Erlang expressions can be used as guard expressions;
+% in particular, our `is_cat` and `is_dog` functions cannot be used within the
+% guard sequence in `is_pet`'s definition.
```
vendethiel: is there a link to all the BIFs?

Jubobs: A comprehensive list can be found in this [section](http://erlang.org/doc/reference_manual/expressions.html#id81912) of the Erlang reference manual. Should I add that link to that page in `erlang.html.markdown`?

vendethiel: I think it'd be nice :).

Jubobs: I'll modify my pull request accordingly. Sit tight :) Incidentally, I've got `[erlang/fr]` lined up after this.

vendethiel: I'll be glad to review that one :)

Jubobs:
I see that my pull request hasn't been accepted, yet. Is anything missing? If so, please let me know.

erlang.html.markdown
```diff
@@ -167,8 +167,15 @@ is_dog(A) -> false.
...
 % by semicolons (`;`). The guard sequence `G1; G2; ...; Gn` is true if at
 % least one of the guards `G1`, `G2`, ..., `Gn` evaluates to `true`.
-is_pet(A) when is_dog(A); is_cat(A) -> true;
-is_pet(A) -> false.
+is_pet(A) when is_atom(A), (A =:= dog) or (A =:= cat) -> true;
+is_pet(A)                                             -> false.
+
+% Warning: not all valid Erlang expressions can be used as guard expressions;
+% in particular, our `is_cat` and `is_dog` functions cannot be used within the
+% guard sequence in `is_pet`'s definition. For a list of the built-in
```
geoffliu: Nitpick: this makes it sound like only BIFs are allowed in guards. How about "For a description of expressions allowed in guard sequences..."? Or "specification"?

Jubobs: You're right; as it is, the passage is a bit misleading. I'll make a new commit, for clarification.

vendethiel:
Thanks!

PR #1107, opened 2015-05-20
[haxe/en] Few cosmetic fixes

sm4rk0:
* correct the array name in a for loop iterating on an index
* add an assignment for "case _" in first switch example
* fix case of variable name in trace for control statement returning as an expression
* move some inline comments to a separate line above the statement and shorten some comments to prevent text overflow

### DIFF
--- a/haxe.html.markdown
+++ b/haxe.html.markdown
@@ -340,7 +340,7 @@ class LearnHaxe3{
         // (more on ranges later as well)
         var n = ['foo', 'bar', 'baz'];
         for (val in 0...n.length){
-            trace(val + " is the value for val (an index for m)");
+            trace(val + " is the value for val (an index for n)");
         }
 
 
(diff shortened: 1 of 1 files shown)


haxe.html.markdown
```diff
@@ -495,8 +495,10 @@ class LearnHaxe3{
         // foo_instance.public_read = 4; // this will throw an error if uncommented:
         // trace(foo_instance.public_write); // as will this.
 
-        trace(foo_instance + " is the value for foo_instance"); // calls the toString method
-        trace(foo_instance.toString() + " is the value for foo_instance.toString()"); // same thing
+         // calls the toString method
```
vendethiel: please add colon for this comment and the next one

haxe.html.markdown
```diff
@@ -534,9 +536,10 @@ class FooClass extends BarClass implements BarInterface{
 
     // a public constructor
     public function new(arg:Int){
-        super(); // call the constructor of the parent object, since we extended BarClass
+        // call the constructor of the parent object, since we extended BarClass
```
vendethiel: same here

vendethiel:
Looks good outside from that :)

haxe.html.markdown
sm4rk0: OK. I actually considered adding a colon, but decided not to do so in order to maintain the same style as the rest of the code, which is mainly without colons.
I hope I'll figure out how to make this changes, as I'm new here...

vendethiel:
Thanks! :cake:

PR #1110, opened 2015-05-22
editorial revert of a change, remove trailing whitespace

tomnor:
Now the regular expression is back to what it was and align with how it is in
other languages. Also it aligns with the commented description of the regular
expression, (`The regular expression is "Bonjour \\(.+\\)!" and it reads:`)


levibostian:
Thanks!

PR #1111, opened 2015-05-23
A bunch of small changes to aid clarity, brevity and/or consistency, …

drguildo:
…none of which should alter the semantics.


levibostian:
Thanks!

PR #1115, opened 2015-05-26
Improve description of the for loop.

drguildo:


geoffliu:
Nice, thanks!

geoffliu:
Actually, on second thought, "iterable" would be a much better word here than "collection."

PR #1117, opened 2015-05-28
Fix typo of chinese translation

ryanouyang:
Line 344: From "行数" to  "函数", which should means 'Function' here.


geoffliu:
Nice catch!

PR #1120, opened 2015-05-30
Use keywords in POST/PUT examples

bobbywilson0:
Using strings doesn't work for the POST/PUT examples here, changing them to keywords fixes this


vendethiel:
Thanks!

PR #1121, opened 2015-06-01
[c++/en] Fix a few coding errors.

adamheins:
Removed the `void` return type from the Dog class constructor and destructor, since constructors and destructors do not have return types. Changed the call to `printDog` to just `print`, matching what was declared in the class. Also added a few missing semicolons.


levibostian:
Thanks! Great catches.

PR #1122, opened 2015-06-01
Added a section on Stages of Git and fixed the add command section

avjinder:


git.html.markdown
```diff
@@ -79,6 +79,11 @@ this pointer will automatically update and point to the latest commit.
 HEAD is a pointer that points to the current branch. A repository only has 1 *active* HEAD.
 head is a pointer that points to any commit. A repository can have any number of heads.
 
+###Stages of Git
```
levibostian: Glad you added this here. Would it be best to switch the "committed" and "modified" and the "staged" and "committed" around to indicate files go from modified to staged to committed?

avjinder: You're right, it should be modified -> staged -> committed. Must have had a brain fart while writing it.

geoffliu:
This is actually one of the most common topics that trip up people who switch from other source control systems, definitely a great addition.

PR #1123, opened 2015-06-02
[c++/en] Make code compileable.

adamheins:
Made the template example class Box have a public member function so it works when that function is called later on. Also removed the string arguments from the std::exception constructors, because std::exception constructors only either take no argurments or another std::exception object.

### DIFF
--- a/c++.html.markdown
+++ b/c++.html.markdown
@@ -519,12 +519,13 @@ printMessage<10>();  // Prints "Learn C++ faster in only 10 minutes!"
 // (see http://en.cppreference.com/w/cpp/error/exception)
 // but any type can be thrown an as exception
 #include <exception>
+#include <stdexcept>
 
 // All exceptions thrown inside the _try_ block can be caught by subsequent
 // _catch_ handlers.
 try {
     // Do not allocate exceptions on the heap using _new_.
-    throw std::exception();
+    throw std::runtime_error("A problem occurred");
 }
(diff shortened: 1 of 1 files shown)


c++.html.markdown
```diff
@@ -523,7 +524,7 @@ printMessage<10>();  // Prints "Learn C++ faster in only 10 minutes!"
 // _catch_ handlers.
 try {
     // Do not allocate exceptions on the heap using _new_.
-    throw std::exception("A problem occurred");
```
vendethiel: this could maybe use `std::runtime_error` instead?

adamheins: Changed to use `std::runtime_error`. Also added the include for it, `#include <stdexcept>`.

c++.html.markdown
```diff
@@ -614,7 +616,7 @@ void doSomethingWithAFile(const char* filename)
 {
     FILE* fh = fopen(filename, "r"); // Open the file in read mode
     if (fh == nullptr)
-        throw std::exception("Could not open the file.");
+        throw std::exception();
```
vendethiel: Can you change this one as well? I'll merge right after. Thanks!

adamheins: Done.

vendethiel:
Thanks :+1:

vendethiel:
:cake:

PR #1125, opened 2015-06-06
Add record notes to the Standard ML page

Inaimathi:
Tripped over this a little while ago when seeing a record in the wild for the first time. I think it's worth noting as a compound datatype along with lists and tuples.


geoffliu:
Nice, thanks!

Issue #562, opened 2014-03-09
[request] D


amilios:
I may be able to work on this one later.

Facebook is starting to use D in production! It is an interesting language.

oldmud0:
Tomorrow night is here!

NickPapanastasiou:
Oh dear, this is embarrassing. I actually think I have something lying around. I'll take a peek through my mess after work tomorrow.

PR #1134, opened 2015-06-10
D language!

NickPapanastasiou:
Here's a little bit of D for ya! Not complete yet, but a solid point for others to jump in if they'd like. I'll of course keep adding to it.

And thanks to @oldmud0 for reminding me that I took on #562.


d.html.markdown
```diff
@@ -0,0 +1,242 @@
+---
+language: D 
+filename: learnd.d 
+contributors:
+    - ["Nick Papanastasiou", "www.nickpapanastasiou.github.io"]
+lang: en
+---
+
+If you're like me and spend way too much time on the internet, odds are you've heard 
+about [D](http://dlang.org/). The D programming language is a modern, general-purpose,
```
geoffliu: Nit: this sentence is suuuuuuper long.

d.html.markdown
```diff
@@ -0,0 +1,242 @@
...
+    while(n < 10_000) {
+        n += n;
+    }
+
+    do {
+        n -= (n / 2);
+    } while(n > 0);
+
+    // For and while are nice, but in D-land we prefer foreach
+    foreach(i; 1..int.max) { // The .. creates a continuous range
```
geoffliu: Is the range inclusive on either side?

d.html.markdown
```diff
@@ -0,0 +1,242 @@
...
+import std.stdio;
+
+void main() {
+    for(int i = 0; i < 5; i++) {
+        writeln(i);
+    }
+
+    auto n = 1; // use auto for type inferred variables
+
+    while(n < 10_000) {
```
geoffliu: Can numeric literals be written without the underscore?

NickPapanastasiou: @geoffliu Yes they absolutely can, I'll add an example without it so it's clear.

d.html.markdown
```diff
@@ -0,0 +1,242 @@
...
+// Here, T is a type parameter. Think <T> from C++/C#/Java
+struct LinkedList(T) {
+    T data = null;
+    LinkedList!(T)* next; // The ! is used to instaniate a parameterized type. Again, think <T> 
+}
+
+class BinTree(T) {
+    T data = null;
+
+    BinTree!T left;
```
geoffliu: There are parentheses around the `T` in the `LinkedList` instantiation, but not here. Are they optional? Also, what does the syntax look like with two type parameters?

d.html.markdown
NickPapanastasiou: Ranges built from a..b are beginning-inclusive, end-exclusive, [a, b). So 1..5 == [1, 2, 3, 4]

d.html.markdown
```diff
@@ -0,0 +1,242 @@
...
+
+T max(T)(T a, T b) {
+    if(a < b) 
+        return b;
+
+    return a;
+}
+
+// Use the ref keyword to pass by referece
+void swap(T)(ref T a, ref T b) {
```
geoffliu: You mentioned up top that structs and unions are passed by value, and classes, by reference. Does the `ref` keyword override that behavior? Or is it because this method is parameterized, so you don't know which you're getting? That is, if `T` here is a struct, is it still passed by reference?

d.html.markdown
```diff
@@ -0,0 +1,242 @@
...
+
+class MyClass(T, U) {
+    T _data;
+    U _other;
+
+}
+
+// We define "setter" methods as follows
+
+class MyClass(T, U) {
```
geoffliu: Let's get rid of this definition; the next one is a strict superset.

d.html.markdown
NickPapanastasiou: Exactly. If T is a value type, and a value of type T, say t, is passed to a function with a parameter of type ref T, then t is passed by reference.

d.html.markdown
```diff
@@ -0,0 +1,242 @@
...
+    writeln(mc.data);
+    writeln(mc.other);
+}
+```
+
+With properties, we can add any amount of validation to
+our getter and setter methods, and keep the clean syntax of
+accessing members directly!
+
+Other object-oriented goodness for all your enterprise needs
```
geoffliu: Eh, I wouldn't call these features "enterprise." It's just good OOP practice.

d.html.markdown
```diff
@@ -0,0 +1,242 @@
...
+
+```d
+import std.algorithm;
+
+void main() {
+    // We want to print the sum of a list of squares of even ints
+    // from 1 to 100. Easy!
+
+    // Just pass lambda expressions as template parameters!
+    auto num = iota(1, 101).filter!(x => x % 2 == 0)
```
geoffliu: What's `iota`? Also, what's the syntax to declare a function type?

d.html.markdown
NickPapanastasiou: It was mainly a dig at Java. If the joke is distracting, we can remove it.

d.html.markdown
NickPapanastasiou: We can declare function types as `returnT function(ArgT1, ArgT2, ...)` or `returnT delegate(ArgT1, ArgT2, ...)`, but lambdas are type-inferred, and one does not generally 
add the type declarations in D. I feel it's safe to omit this.

d.html.markdown
```diff
@@ -0,0 +1,242 @@
...
+                           .map!(y => y ^^ 2)
+                           .reduce!((a, b) => a + b);
+
+    writeln(num);
+}
+```
+
+Notice how we got to build a nice Haskellian pipeline to compute num?
```
geoffliu: Hmm this is really interesting, but raises a few questions. Suppose you have a method `foo` on class `Bar` and an instance `bar` of `Bar`. Is `bar.foo(...)` the same as `Bar.foo(bar, ...)`?

Also, is it possible, in D, to pass functions or methods as first-class values? There are lambdas, so I'm guessing functions are fair game (a little explanation of it would be nice though). What about methods? Can I pass `bar.foo` as a value? Does that automatically get turned into a partially applied function?

vendethiel:
Good to see this :)

geoffliu:
Complete newbie to D, so comments are just what I thought was confusing/ambiguous as a newcomer, and I suppose others will have the same confusion.

Thanks for the write up though, D looks like a really interesting language. Look forward to building this page out a little more, especially with some of the low-level features you mentioned. For now, however, let's get some of the confusing points taken care of and merge this one. :)

d.html.markdown
geoffliu: Ha, I hear ya. I only-ever-so-slightly hate Java as the next guy, but this one did catch me by surprise. I guess it wasn't obvious enough that this is a joke. Let's remove it.

d.html.markdown
geoffliu: I guess my point was that what if I need to define a function like `filter`? What would the signature look like?

NickPapanastasiou: Ooh, that's a whole can of worms there... You see, D's templates are fantastic and ultimately easy to use, but type signatures can be a little absurd. From the www.dlang.org documentation,
we have the type signature of `filter` as instatiated here to be
 `auto filter(alias predicate, Range)(Range rs) if (isInputRange!(Unqual!Range)) && is(typeof(unaryFun!predicate));` 

Once you're eyes are done bleeding, it's not so bad, but it takes some getting used to. We can read this as "auto: return type is inferred", "(alias predicate, Range): predicate is some symbol available at compile time and Range is any type. Neat.", " if (isInputRange!(Unqual!Range)) && is(typeof(unaryFun!predicate)): we restrict the type of Range to any type that is an InputRange, and preductate must be a unary function. Got it." This is a bit much for 
learnxinyminutes, so we should probably just hand wave the "type sorcery" as we call it.

d.html.markdown
NickPapanastasiou: No, UFCS is applied to free functions to appear as though they are methods, but not the other way around. So if `foo` is a method on `Bar` and `bar` is an instance of `Bar`, we may not write `foo(bar)` to replace `bar.foo()`.

2. Yes! We can even go point-free and write

```d

auto num = iota(1, 101).filter!(isEven)
                                     .map!(square)
                                     .reduce!(add);
```

assuming the natural functions are defined. Partial application does not happen by defualt, though. See `std.functional` for the `partial` template for this functionality (see what I did there?)

d.html.markdown
geoffliu: OK, let's add a note there then. Also, does that work on primitive types as well? If `T` happens to be `int`, is that still by ref?

d.html.markdown
geoffliu: Let's add a note about it. We wouldn't have to do this if it weren't for Matlab and coffeescript...

d.html.markdown
NickPapanastasiou: Yes, primitives are value types and hence copied by default. The `ref` qualifier in a parameter's definition would then ensure that any type (be it reference or value) is passed by reference.

d.html.markdown
geoffliu: OK, let's do a quick note about 1 and 2 (the part where you pass functions directly).

Re: partial application: can you refer to a method as a function then? As in, if I write `.filter(bar.foo)`, is the `bar` instance bound there? Or do we need extra syntax to make it work?

d.html.markdown
geoffliu: I... I see. Interesting way to incorporate type classes. OK let's leave that out.

d.html.markdown
NickPapanastasiou: Note that you'd write `.filter!(bar.foo) // The function is a template parameter`. And no, if it's a method, that's all it is. There is no way to rewrite method calls as free function calls in D.

d.html.markdown
```diff
@@ -0,0 +1,226 @@
...
+module hello;
+
+import std.stdio;
+
+// args is optional
+void main(string[] args) {
+    writeln("Hello, World!");
+}
+
+// Conditionals and loops work as expected.
```
vendethiel: maybe move this before the "for"?

d.html.markdown
```diff
@@ -0,0 +1,226 @@
...
+```
+
+Then we may write 
+
+```d
+A.f(B, C, ...)
+```
+
+and the two are equivalent! No more fiddling to remember if it's
+str.length or length(str)!
```
vendethiel: Would you mind adding *some* references here? website and al. thanks

d.html.markdown
NickPapanastasiou: D's templates can actually be viewed as a strict generalization of type classes, with worse syntax. You'll see layer that their metaprogramming power makes this a sensible tradeoff.

geoffliu:
@NickPapanastasiou gentle ping. Could you make the comments still outstanding, and we can get this merged?

NickPapanastasiou:
Sorry bud, work kind of smacked me the past few days. I'll get them in the morning!

oldmud0:
404s on the website.

PR #1145, opened 2015-06-16
[Ruby/en] Fix explanation of "and" and "or"

### LINKED ISSUE #1144: Ruby logical operators
The logical operators ```&&``` and ```||``` are for Boolean logic while the operators ```and``` and ```or``` are for control flow. They have a completely different precedence and will group things wrong if you use the ```and``` and ```or``` syntax without parenthesis. 

for a deeper explanation here:

http://devblog.avdi.org/2010/08/02/using-and-and-or-in-ruby/

geoffliu:
Fixes #1144.


vendethiel:
:cake:

PR #1149, opened 2015-06-19
Fix to "if" conditional in bash docs

ilan1:


vendethiel:
Whoops, seems like something went wrong still :)
you should probably rebase your PR (and maybe force-push)

PR #1150, opened 2015-06-19
Fixed misinfo regarding sigils

a3f:
Variables might start with `$`. Edited to make that point clearer.
Also `#` is more commonly called number sign than number symbol.


vendethiel:
i'm actually afraid of using "type" here :-).

a3f:
Sorry for the later answer.
@vendethiel The official documentation calls them data types too.

`perldoc perldata`:
> Perl has three built-in data types: scalars, arrays of scalars, and associative arrays of scalars, known as "hashes".

`perldoc perlguts`:
> Perl has three typedefs that handle Perl's three main data types:

vendethiel:
Alright, thanks. That seems good.

PR #1151, opened 2015-06-22
Remove in-development from Rust's description

mjs2600:
Now that Rust is stable at 1.0 saying "in-development" might not be appropriate any more.


vendethiel:
Nice catch! :cake: (<- that's the Rust 1.0 cake!)

PR #1159, opened 2015-07-05
python/en - added two resources

mjhea0:


levibostian:
Thanks for your contribution! 

I enjoy the first link you included but the root link to realpython.com is a link to a paid service which we do not accept into this repo. So if you could remove that link that would be great.

mjhea0:
updated! thanks!

levibostian:
Thanks!

PR #1165, opened 2015-07-15
Chapel Tutorial

ian-bertolacci:
Created a from-scratch tutorial for the Chapel Language.
Code compiles as seen using release version Chapel 1.11.0
Covers many topics that a beginner would be interested in knowing but not interested enough to scour the specification, miscellaneous slides, or technical academic and white papers.

### DIFF
--- a/chapel.html.markdown
+++ b/chapel.html.markdown
@@ -49,7 +49,7 @@ hear about what fiber mixture the ropes are, or how they were braided, or how th
 differ between one another.
 It won't teach you how to develop amazingly performant code, and it's not exhaustive. 
 Refer to the [language specification](http://chapel.cray.com/language.html) 
-and the [library-documentation](http://chapel.cray.com/docs/latest/) for more details.
+and the [library documentation](http://chapel.cray.com/docs/latest/) for more details.
 
 Occasionally check here back to see if more topics have been added.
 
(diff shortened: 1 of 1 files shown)


adambard:
Neat! I don't think chapel is covered by Rouge, so I'm going to see what c highlighting ends up looking like. It won't be perfect, but at least comments won't look like the rest.

Thanks!

ian-bertolacci:
Interesting. 
It didnt particularly like my lists at the bottom of the page, even though the github markdown preview did.
I'll see if I can fix that at some point.

Also, as for syntax highlighting, C works.
Is there a way to add our own? 
Chapel comes with source-highlight, emacs, and vim highlighting codes.

adambard:
The syntax highlighting is from [my fork of rouge](https://github.com/adambard/rouge) for silly reasons related to version number compatibility, but it tracks the [main rouge project](https://github.com/jneen/rouge) as far as it needs to so you could definitely contribute there if you like (or to the fork if you come up with something too half-assed for real production).

The lists look ok to me, but feel free to open a PR with any edits you have!

ian-bertolacci:
Since I'm trying to do other chapel syntax highlighting for chapel as well, I might as well include Rouge somewhere.

As for the lists, section 'Installing the Compiler' has a numbered list that seems to have been flattened (ie 1. item 2. item ) and section 'Compiling Code' has a bulleted list of one element.
Which is interesting because the bullet list in the preceeding 'What this tutorial is lacking' section looks fine.
I say this from looking at it with Firefox and Safari from OS X Yosemite 
Its no biggie anyhow. I'm sure people will figure it out.

PR #1167, opened 2015-07-16
Changed %s formatting to {n} style.

MrMallIronmaker:
If line 117 is accurate, then {n} style string formatting is the preferred style, and it should be reflected in the examples.

### DIFF
--- a/python.html.markdown
+++ b/python.html.markdown
@@ -327,8 +327,8 @@ prints:
     mouse is a mammal
 """
 for animal in ["dog", "cat", "mouse"]:
-    # You can use % to interpolate formatted strings
-    print "%s is a mammal" % animal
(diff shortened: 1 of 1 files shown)


adambard:
I think the {n} style should be preferred in general, but % formatting should definitely be at least mentioned. Who knows, maybe the reader will someday have to read Python code instead of just writing it!

MrMallIronmaker:
Oops, sorry about the close / reopen, that's a stray finger on my phone.

I didn't take out the original referemce to 
%s formatting, (like you said, they may as well know it for reading) but just changed any references farther down in the code.

adambard:
Ah, well in that case, sounds good to me. I'll ping @louiedinh about it because I know he likes to know about this stuff, but it's definitely the right way.

Thanks for the contribution!

PR #1169, opened 2015-07-18
Merged fix from upstream, added slight corrections

ian-bertolacci:
After down merging, included correct capitalization of Chapel, correct use of apostrophes.

### DIFF
--- a/chapel.html.markdown
+++ b/chapel.html.markdown
@@ -1,11 +1,11 @@
 ---
-language: chapel
+language: Chapel
 filename: learnchapel.chpl
 contributors:
     - ["Ian J. Bertolacci", "http://www.cs.colostate.edu/~ibertola/"]
 ---
 
-You can read all about chapel at [Cray's official Chapel website](http://chapel.cray.com).
+You can read all about Chapel at [Cray's official Chapel website](http://chapel.cray.com).
 In short, Chapel is an open-source, high-productivity, parallel-programming language in development at Cray Inc., and is designed to run on multi-core PCs as well as multi-kilocore supercomputers.
 
 More information and support can be found at the bottom of this document.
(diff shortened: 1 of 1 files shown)


chapel.html.markdown
```diff
@@ -1,11 +1,11 @@
 ---
-language: chapel
-filename: learnchapel.chpl
+language: Chapel
```
levibostian: This should be changed back to "chapel" as it will break the site more then likely.

chapel.html.markdown
```diff
@@ -1,11 +1,11 @@
...
+filename: learnChapel.chpl
```
levibostian: This should be changed back to "learnchapel.chpl" as it will break the site more then likely.

chapel.html.markdown
```diff
@@ -201,8 +201,6 @@ select( inputOption ){
   otherwise { 
     writeln( "Any other Input" );
     writeln( "the otherwise case doesn't need a do if the body is one line" );
-    writeln( "Oh, and when statements dont cascade like the case statements" );
```
levibostian: Why did you decide to remove these 2 lines?

chapel.html.markdown
ian-bertolacci: Accepted.

chapel.html.markdown
ian-bertolacci: Accepted.

chapel.html.markdown
ian-bertolacci: I felt it was redundant given the comment above
"However, Select statements don't cascade like in C or Java"

It may not be necessary to remove output.

levibostian: Got ya. :+1:

levibostian:
Thanks!

Issue #1148, opened 2015-06-19
Use next() built-in for Python 3


wbolster:
The Python 3 page should use `next(iterator)` instead of `iterator.__next__()`. The `next()` function is a built-in which under the hood will call `x.__next__()`, but the latter is considered an implementation detail of the iterator protocol, while the `next()` function provides a public interface to advance an iterator.  (Similar for `len(x)` versus `x.__len__()`, and so on.)

vendethiel:
Nice catch! Would you mind sending a pull request our way fixing that :-)?

PR #1175, opened 2015-07-24
[python3/en] Use `next()` instead of `__next__()`

genkimarshall:
Fixes issue #1148. The reasoning is well explained by the issue.
One can also refer to the [docs], showing it is more idiomatic to use `next()`.

[docs]: https://www.python.org/dev/peps/pep-3114/#double-underscore-methods-and-built-in-functions


geoffliu:
:+1:

PR #1177, opened 2015-08-02
Chapel: Sync, atomics, reductions, scans, zippered iterators

ian-bertolacci:
Added sync, single and atomic variables.
Added reductions and scans.
Because max/minloc require the values and indices to be in a zippered iterator, I included a demonstration of zippered iterators along with the previous iterator example. 

Further, made a number of spelling, grammar, wording, punctuation changes, and corrected/clarified some erroneous/misleading comments (mainly things about consts and params)

Included the `--set` compiler option in the Notable arguments 'section'.

### DIFF
--- a/chapel.html.markdown
+++ b/chapel.html.markdown
@@ -765,6 +765,42 @@ proc countdown( seconds: int ){
   }
 }
 
+// Atomic variables, common to many languages, are ones whose operations
+// occur uninterupted. Multiple threads can both modify atomic variables
+// and can know that their values are safe.
+// Chapel atomic variables can be of type bool, int, uint, and real.
+var uranium: atomic int;
+uranium.write( 238 ); // atomically write a variable
+writeln( uranium.read() ); // atomically read a variable
+// operations are described as functions, you could define your own operators.
+uranium.sub( 3 ); // atomically subtract a variable
+writeln( uranium.read() );
+var replaceWith = 239;
+var was = uranium.exchange( replaceWith ); 
+writeln( "uranium was ", was, " but is now ", replaceWith );
+var isEqualTo = 235;
+if uranium.compareExchange( isEqualTo, replaceWith ) {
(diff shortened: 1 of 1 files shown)


vendethiel:
Nice work, thanks! Seems good, except for "zippered", I've never seen that name before. Are chapel's docs using this word?

ian-bertolacci:
Talking with the developers they call such iterations zippered. 
Looking in the texted it can be somewhat varied. 
Section 11.8.1 calls it Zipper Iteration, and 18.4.3 calls it Zippered Iteration.

I think both work in the phrasing zipper(ed) iterators and zipper(ed) iteration.

vendethiel:
would you mind putting `(zipped iterators)` (in parentheses) next to the word? I'll merge it then, thanks :-).

ian-bertolacci:
I like that idea. However I think I misspoke in the text. I dont believe zip() itself is an iterator; not that this changes much.

What I want to say now is:
We can zipper two or more iterators (who have the same number of iterations) together using zip() to create a single zipped iterator, where each iteration of the zipped iterator yields a tuple of the values yielded by one iteration each of the included iterators.

Does this help clear things up? or is more complicated now? 
I think I will also change the example so that the two iterators are quite different and the source of the iteration values could not be confused.

vendethiel:
> yields a tuple of the values yielded by one iteration each of the included iterators.

I think it'd be cleared with "yields a tuple of one value yielded from each iterator" or something like this. That's the only "confusing" part. I'd agree with the remaining one :).

ian-bertolacci:
Agreed. I like that wording better.

vendethiel:
Thanks for your contribution! :cake: :fish_cake:

ian-bertolacci:
No prob! Thanks for including the work!
It makes it easier when I tell people to go learn Chapel if there is actually a place where they could learn Chapel.

PR #1178, opened 2015-08-02
Fix 'filename' collision between csharp{,-tr}.html.markdown

thiagovandieten:
Both were set to LearnCSharp.cs, so the live site has been serving the Turkish version for English.


vendethiel:
Thanks!

PR #1180, opened 2015-08-04
Println deprecated in Swift 2, added name

anthonyn60:
In Swift 2, println will no longer work. Changed all println's to print, for compatibility with both Swift 2 and Swift 1.2.


vendethiel:
Looks good, thanks! :ship:

PR #1183, opened 2015-08-06
[Chapel] Modules, main(), range/domain/array slicing and array assignment, loop expressions, zipped iterators

ian-bertolacci:
This is quite the doozie.
Added modules and main() proc.
Added range/domain/array slicing.
Added array assignment (which is part of array slicing) and loop expressions, which necessitated a deeper dive into zipped iterators to explain why some expressions would work, and why some wouldn't.
Added a fix because `<THESE THINGS>` were missing from the Notable arguments section.
A small change was made in one or two places to keep within the 80 column specifications.

### DIFF
--- a/chapel.html.markdown
+++ b/chapel.html.markdown
@@ -24,7 +24,7 @@ writeln( "World!" );
 // each thing is printed right next to each other, so include your spacing!
 writeln( "There are ", 3, " commas (\",\") in this line of code" );
 // Different output channels
-stdout.writeln( "This goes to standard output (just like plain writeln( ) does)");
+stdout.writeln( "This goes to standard output, just like plain writeln() does");
 stderr.writeln( "This goes to standard error" );
 
 // Variables don't have to be explicitly typed as long as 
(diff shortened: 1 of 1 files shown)


chapel.html.markdown
```diff
@@ -310,6 +311,18 @@ stringSet += "a"; // Redundant add "a"
 stringSet -= "c"; // Remove "c"
 writeln( stringSet ); 
 
+// Both ranges and domains can be sliced to produce a range or domain with the
+// intersection of indices
+var rangeA = 1.. ; // range from 1 to infinity
+var rangeB =  ..5; // range from negative infinity to 5
```
levibostian: This is nothing wrong with your pull request, it is just more my curiosity. "negative infinity" I wonder where it *actually* starts? It can't be negative infinity, can it?

levibostian:
Looks great. Thanks!!

chapel.html.markdown
ian-bertolacci: So per the spec (section 18.2) the 'boundness' of the range is stored as an enum of bounded, boundedLow, boundedHigh, boundedNone.
Im not sure what the implementation is (though Im going to ask around now) but If one were to 
```c
for i in max(int(8))-1.. do something(i);
```
it will just overflow to min(int(8)) and keep going in an infinite loop.
I guess infinity means a lot of things.

ian-bertolacci: That said you cannot iterate over an range that is unbounded on its low side, since it doesnt know where to start. 
Thats a run-time error.

levibostian: :+1: Thank you! I had a feeling it had to be something like that, I just was super curious.

PR #1193, opened 2015-08-08
added a scratch for E4 language

Quasilyte:
Maybe it is too early to post anything,
but it is so damn tempting! Love LXinY.

### DIFF
--- a/e4.html.markdown
+++ b/e4.html.markdown
@@ -1,0 +1,228 @@
+---
+language: E4
+contributors:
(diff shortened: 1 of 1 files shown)


levibostian:
I don't consider this too early, it has some good content. Are you planning on working on it a little while longer and then making another pull request?

Quasilyte:
Yeah, I need another few days.

levibostian:
:+1:

PR #1194, opened 2015-08-08
E4 -- Emacs Forth language

Quasilyte:


vendethiel:
(If you push new commits to your branch, github will pick them up. I can't properly review ATM because I'm on holidays, but I'll do so next week if possible)

PR #1197, opened 2015-08-10
update "further reading" links

dander:
Change the chat room reference from Stack Overflow chat to Gitter.im, and add a link to the "Red" Stack Overflow tag.


levibostian:
Thanks!

PR #1199, opened 2015-08-14
[Perl/en] translated to [Perl/pt-br]

miguelarauj1o:


adambard:
Can you make the `filename` field in the header language-specific? `learnperl-pt.pl` should work fine. Otherwise the files will conflict with english because the site backend is kind of janky that way.

miguelarauj1o:
ooops, sorry :)

PR #1204, opened 2015-08-18
Add example of curly syntax for variables

casual-passer:


vendethiel:
perhaps the previous example should be changed instead?

vendethiel:
@casual-passer ?

PR #1205, opened 2015-08-22
[Swift/tr] Turkish For Swift

ozgurshn:
[Swift/tr] Turkish For Swift


vendethiel:
Thanks! I'll just leave that open for a bit if someone else want to review it

PR #1210, opened 2015-08-25
In perl6, 0 is not falsey anymore

AlexDaniel:
0 is not falsey for a long time. For example, see this blog post: https://p6weekly.wordpress.com/2015/04/13/2015-15-we-figured-out-the-truth-about-0/

> Something that may have surprised a lot of people (myself included) was that Perl 6 claims that the truth value of “0” ought to be False. That is no longer the case, and so the rule for the truth value of strings is now just “does it have any characters in it?”. I think that’s more sensible.

### DIFF
--- a/perl6.html.markdown
+++ b/perl6.html.markdown
@@ -213,7 +213,7 @@ say $x; #=> 52
 # - `if`
 # Before talking about `if`, we need to know which values are "Truthy"
 #  (represent True), and which are "Falsey" (or "Falsy") -- represent False.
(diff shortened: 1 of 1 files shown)


vendethiel:
Nice catch, thanks! :+1:

AlexDaniel:
Actually, I've mistaken ``0`` with ``"0"``. This is fixed in https://github.com/adambard/learnxinyminutes-docs/commit/62f27af4c38f546f29fcf095fc653989b0667908

PR #1212, opened 2015-08-26
Add #push to Array besides shovel operator

arthur-vieira:
#push is commonly used imho and it is also used in other languages (e.g. Javascript).


levibostian:
Thanks!

PR #1213, opened 2015-08-26
[tcl/en] Updating line 167 to fix name resolution error at line 173. 

asfarley:
Also, elaborated on name resolution with set vs. variable.


vendethiel:
:+1:

PR #1221, opened 2015-09-01
Tcl. Fix mid-page rendering error

josteink:
Github's syntax-highlighting has an error in the Tcl syntax-parser causing the rendering of the page to fail about one third in.

See the following screenshot:

error-example

This commit bypasses this rendering-error by resetting the syntax highlighting after the error has occurred.


vendethiel:
... Fair enough. We probably should open an issue though...

PR #1222, opened 2015-09-01
Python: add finally and with statements

outcoldman:
https://docs.python.org/2/tutorial/errors.html

### DIFF
--- a/python.html.markdown
+++ b/python.html.markdown
@@ -382,7 +382,7 @@ else:   # Optional clause to the try/except block. Must follow all except blocks
 finally: #  Execute under all circumstances
     print "We can clean up resources here"
 
-# Instead of try/finally to cleanup resources you can use with
+# Instead of try/finally to cleanup resources you can use a with statement
 with open("myfile.txt") as f:
     for line in f:
         print line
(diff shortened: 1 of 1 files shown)


python.html.markdown
```diff
@@ -379,7 +379,13 @@ except (TypeError, NameError):
     pass    # Multiple exceptions can be handled together, if required.
 else:   # Optional clause to the try/except block. Must follow all except blocks
     print "All good!"   # Runs only if the code in try raises no exceptions
+finally: #  Execute under all circumstances
+    print "We can clean up resources here"
 
+# Instead of try/finally to cleanup resources you can use with
```
geoffliu: Instead of "with", could you say "a with statement"? Or enclose "with" in backticks.

geoffliu:
Also while we're at it, could you update the Python3 page as well? Thanks!

python.html.markdown
outcoldman: fixed!

outcoldman:
@geoffliu of course, but I'm too lazy to enlist this repo, so created a separate pull request using GitHub UI

geoffliu:
Thanks!

PR #1223, opened 2015-09-01
Python3: add finally and with statements

outcoldman:


geoffliu:
Thanks for updating this one too!

PR #1225, opened 2015-09-02
[brainfuck/pl] translated from en

hckr:


adambard:
We normally leave stuff up for review, but I can't imagine that a mistake in brainfuck is going to bring down anyone's server at 3am. Bardzo dziękuję!

PR #1226, opened 2015-09-02
[javascript/en] described briefly falsy values, strict mode, treating…

hckr:
… strings as arrays and added an example of short-circuit and

### DIFF
--- a/javascript.html.markdown
+++ b/javascript.html.markdown
@@ -3,6 +3,7 @@ language: javascript
 contributors:
     - ["Adam Brenecki", "http://adam.brenecki.id.au"]
     - ["Ariel Krakowski", "http://www.learneroo.com"]
+    - ["Jakub Młokosiewicz", "https://github.com/hckr"]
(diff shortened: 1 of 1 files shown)


javascript.html.markdown
```diff
@@ -141,6 +144,14 @@ someOtherVar = 10;
 // ...but your variable will be created in the global scope, not in the scope
 // you defined it in.
 
+// ...but if you enter strict mode, after using the following statement:
+
+"use strict"; // or 'use strict'
```
vendethiel: is everything after that line correct strict mode?

hckr:
No, it's not correct strict mode, it's not correct even out of strict mode. For example there are places where non-existent functions are called (e.g. doStuff() at the beginning).

adambrenecki:
I'm -1 on explaining truthiness and short-circuiting - I think most of the target audience will know what those things are already. +1 on strict mode, subscripting strings, and adding `var` where it was missing.

PR #1234, opened 2015-09-14
[make/en] Adding a tutorial on (gnu) make

robochat:
Here is a first draft of a make tutorial. I'm not an expert at make at all so corrections are probably needed!

### DIFF
--- a/make.html.markdown
+++ b/make.html.markdown
@@ -111,12 +111,19 @@ small/%.png: %.svg
 %.png: %.svg
 	@echo this rule is chosen
 	
-# however make will use the first pattern rule that can make the target
+# However make will use the first pattern rule that can make the target
 %.png: %.ps
 	@echo this rule is not chosen if %.svg and %.ps are both present
 
 # make already has some pattern rules built-in. For instance, it knows 
 # how to turn *.c files into *.o files.
+
+# Older makefiles might use suffix rules instead of pattern rules
+.png.ps:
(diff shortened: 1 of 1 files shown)


vendethiel:
Looks like a good idea... And there's a lot of ground to cover, as well!

make.html.markdown
```diff
@@ -0,0 +1,236 @@
+---
+language: make
+contributors:
+    - ["Robert Steed", "https://github.com/robochat"]
+filename: Makefile
+---
+
+A Makefile defines a graph of rules for creating a target (or targets).
+It's purpose is the do the minimum amount of work needed to update a
```
vendethiel: Its*
to do*

robochat: I found another it's error too. Embarrassing.

switchhax: Well it´s always going to be reviewed, otherwise you can fix errors also later :)

geoffliu:
I forget, does the GNU variant support, for example, `.c.o` rules?

robochat:
I think that it's supported but I think that they suggest that pattern rules should be used in preference.

geoffliu:
I think it's still worth a mention. I've definitely seen those rules in the wild before, so I can imagine someone needing to know what those are. Nothing extensive, even if you just mention the name and what the look like, so someone has a word to Google search if they need to.

geoffliu:
Thanks, this looks great!

robochat:
Great!

PR #1244, opened 2015-09-22
fix two small bugs - tab error and quotation marks

robochat:
I just found two small errors in the tutorial. One was a tab error!

### DIFF
--- a/make.html.markdown
+++ b/make.html.markdown
@@ -97,7 +97,7 @@ process: ex1.txt file0.txt
 # Can teach make how to convert certain files into other files.
 
 %.png: %.svg
-	inkscape --export-png %.svg
+	inkscape --export-png $^
 
 # Pattern rules will only do anything if make decides to create the \
 target.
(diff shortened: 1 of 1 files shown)


make.html.markdown
```diff
@@ -222,7 +222,7 @@ report:
 ifeq ($(sport),tennis)
 	@echo 'game, set, match'
 else
-	@echo 'They think it's all over; it is now'
+	@echo "They think it's all over; it is now"
```
vendethiel: is that not correct? the echo 2 lines earlier uses single quotes as well

robochat: The ' in  "it's" closes the quotation and then the ";" finishes the statement and then " it is now" throws an error.

robochat:
These errors in the pattern rules are probably more serious. I thought that this was possible to do but it doesn't work on my current computer and I can no longer find any reference that says that it should work like that. So I've changed to using automatic variables, luckily that section is before the pattern section and so it's a logical progression.

make.html.markdown
vendethiel: oh my, sorry for being blind! good catch

vendethiel:
:+1:

PR #1246, opened 2015-09-23
Update perl6.html.markdown

kjkuan:


vendethiel:
Oh, nice catch.

vendethiel:
`[]` might probably make it into the list in time. I just need to get to GLR stuff in !

PR #1250, opened 2015-09-28
[Crystal/en] Crystal language

veelenga:
It would be awesome to be able to learn [Crystal](http://crystal-lang.org/) langauge in Y minutes too :smiley_cat: .

WIP but it's better to get feedback earlier.

### DIFF
--- a/crystal.html.markdown
+++ b/crystal.html.markdown
@@ -21,6 +21,7 @@ true.class #=> Bool
 !0     #=> false
 
 # Integers
+
 1.class #=> Int32
 
 # Four signed integer types
@@ -49,6 +50,7 @@ true.class #=> Bool
 0xfe012d #=> 16646445
 
 # Floats
+
 1.0.class #=> Float64
 
 # There are two floating point types
@@ -60,6 +62,7 @@ true.class #=> Bool
 1.5e-7.class  #=> Float64
 
 # Chars
+
 'a'.class #=> Char
 
 # Octal codepoint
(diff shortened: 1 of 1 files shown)


crystal.html.markdown
```diff
@@ -0,0 +1,164 @@
+---
+language: crystal
+filename: learncrystal.rb
```
vendethiel: why `.rb`? the triple-``` ` ``` should be enough

onebree:
I think it is important to note how Crystal is derived from Ruby. A little history is always good for newer languages!

veelenga:
Good point. Gonna add this note and finish soon

crystal.html.markdown
```diff
@@ -0,0 +1,345 @@
...
+tmp["foo"] = "bar"
+tmp
+
+# Ranges
+
+1..10                  #=> Range(Int32, Int32)
+Range.new(1..10).class #=> Range(Int32, Int32)
+
+# Can be inclusive or exclusive
+3..5.to_a  #=> [3, 4, 5]
```
asterite: Should be `(3..5).to_a`, otherwise it's parsed as `3 .. 5.to_a` (in Ruby too!)

asterite: And also examples below using range should use parentheses

vendethiel:
Seems like it's not being worked on anymore. I'll close this one for now, I'll reopen it if you work on that again :). Thank you!

PR #1257, opened 2015-09-30
[Julia/es] Use smaller images.

Ismael-VC:
I didn't notice they wouldn't fit well, because they fitted ok in gitHub preview.

### DIFF
--- a/es-es/julia-es.html.markdown
+++ b/es-es/julia-es.html.markdown
@@ -9,7 +9,7 @@ filename: learnjulia-es.jl
 lang: es-es
 ---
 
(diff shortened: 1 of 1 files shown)


levibostian:
yeah, that is better. Thanks!

Ismael-VC:
@levibostian thanks! Do you know if there is a way to center the images? I think I can use normal HTML `img` tag, but I wont see the result after the commit is merged and the page rebuilt. That's why I only changed the sizes of the images and used normal markdown syntax.

levibostian:
No, I do not think md alone can do it. You would have to use HTML to center it.

PR #1272, opened 2015-10-01
[c++/pt-br] Fix translate on editing object as parameters

leonnardo:
Fix translate in portuguese about how objects can be edited inside methods.


levibostian:
Thanks!

PR #1287, opened 2015-10-02
[python3/de] Add german translation

matthiaskern:
It's based on the german translation of python2, because of the similarities. 
Because of this, I kept the original translator in the credits.


vendethiel:
thanks&sorry for the delay!

PR #1288, opened 2015-10-02
[python/en] Add note about Python Module Order of Import

DaKnOb:
Just a small note on importing modules for python that can save some debugging time should you have a file with a generic name in a multi-file project.


levibostian:
Please add this to the python3 document as well and add to this PR. Thanks!

vendethiel:
Thanks and sorry for the delay :).

PR #1292, opened 2015-10-02
[PHP] Adds some new php operators

micheleorselli:
This PR adds some new PHP operators, in particular:
- ternary shortcut
- spaceship operator
- null coalesce operator


levibostian:
Thanks!

PR #1295, opened 2015-10-02
[python/en] Remove random "r"

jhuang314:


vendethiel:
Oh, nice catch. I wonder how that one slipped in! Thanks.

PR #1296, opened 2015-10-02
markdown example: fixed missing ! to create an actual comment

edm00se:
Line 163 was missing a character to denote the HTML comment (shown as `<---` instead of `<!---`), which was causing everything subsequently to be rendered as code block, as the triple back tick wasn't being escaped as part of a comment


levibostian:
Nice catch. Thanks!

edm00se:
My pleasure.

PR #1299, opened 2015-10-02
String concatenation with html tags also supported example

dhwanishah:
Added example of when string concatenation can also be helpful with combination of strings with html tags as this is important to understand when templating say output code from a database transaction.

### DIFF
--- a/java.html.markdown
+++ b/java.html.markdown
@@ -1,3 +1,18 @@
+---
+language: java
+contributors:
+    - ["Jake Prather", "http://github.com/JakeHP"]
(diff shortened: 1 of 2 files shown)


vendethiel:
I don't see the point of this. It's just text...

dhwanishah:
The point is to show that in php, you don't have to just concatenate text and variables. Concatenating html tags directly inside the output statements is possible. Rather than ppl just writing html code outside the php code, they can include it directly inside. Not many beginners will try to test this out

PR #1300, opened 2015-10-02
Added how to declare and initialize variables for single and multiple instances

dhwanishah:
Added section on how to declare and initialize both single variable and multiple variables with the same value. Important concept for large structured programs. Separated this a little bit.


levibostian:
Thanks!

PR #1313, opened 2015-10-03
[swift/en] Adding example for for-loop-iteration with array indices

VFUC:


swift.html.markdown
```diff
@@ -147,6 +148,11 @@ for value in myArray {
     }
 }
 
+// for loop with indices (array)
```
levibostian: Did Swift 2.0 change this? With `enumerate` it changed, did this change as well?

Swift 2.0:
```
for (index, item) in myArray.enumerate() {
    print("value: \(item) is in position: \(index)")
}
```

levibostian: I really like `enumerate`. I use it a bit. Get the index, and item both.

VFUC: Yes, I'm pretty sure it changed from `indices(array)` to `array.indices()`
I do agree, enumerate is quite handy and should be included as well.

PR #1324, opened 2015-10-04
Fix usage of "it's" in example comment for chapel

dstockto:


vendethiel:
Thanks! We accept this kind of PR over multiple files, though :)

vendethiel:
:cake:

dstockto:
Thank you, I'll keep that in mind for the future. I didn't want to make it so some could be and others could not be accepted by combining.

PR #1327, opened 2015-10-04
Fix "its" to "it's" in comment for swift-tr doc

dstockto:


vendethiel:
@ozgurshn could you translate this bit please?

dstockto:
Unfortunately, I cannot.

Issue #1170, opened 2015-07-20
{learn c in y minutes } Variable Length Array


xerosanyam:
Here,
printf("Enter the array size: "); // ask the user for an array size
  char buf[0x100];
  fgets(buf, sizeof buf, stdin);

We are already giving the size of array buf.
0x100 in decimal implies 256, so the array is declared to be of size 256, then why are we calling it VLA ?

thePsguy:
Fixing it and submitting a pull request.

PR #1329, opened 2015-10-04
{c}Variable size array, user size input added.

thePsguy:
Fixed Issue #1170
Variable size array, user size input added.

### DIFF
--- a/c.html.markdown
+++ b/c.html.markdown
@@ -130,7 +130,9 @@ int main(void) {
   // can be declared as well. The size of such an array need not be a compile
   // time constant:
   printf("Enter the array size: "); // ask the user for an array size
-  char buf[0x100];
+  int size;
+  scanf("%d", &size);
+  char buf[size];
   fgets(buf, sizeof buf, stdin);
 
   // strtoul parses a string to an unsigned integer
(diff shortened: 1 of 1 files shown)


c.html.markdown
```diff
@@ -130,7 +130,9 @@ int main(void) {
   // can be declared as well. The size of such an array need not be a compile
   // time constant:
   printf("Enter the array size: "); // ask the user for an array size
-  char buf[0x100];
+  int size;
+  scanf("%d", &size);
```
vendethiel: please don't use `scanf`

thePsguy: size needs an integer input. scanf provides a simple proof of concept.
I can't seem to find any other function to take direct integer input. Taking a string and then converting it would be irrelevant to the topic here.
Please tell me if I am wrong?

vendethiel: fgets/sscanf. also see http://stackoverflow.com/questions/1621394/how-to-prevent-scanf-causing-a-buffer-overflow-in-c

thePsguy: Thank you for the reference. fgets takes string, and sscanf can not take input from stdin. Though I stumbled upon fscanf ((fgets + sscanf)? lol) and I think its the best replacement.

vendethiel:
:ship:

thePsguy:
:D

PR #1338, opened 2015-10-04
[ruby/en] Include examples of 'unless' and one-line conditionals

sdulal:
One of Ruby's more unique control structures is the 'unless' statement. This statement was not in the markdown file before.

### DIFF
--- a/ruby.html.markdown
+++ b/ruby.html.markdown
@@ -208,7 +208,9 @@ new_hash.has_value?(3) #=> true
 
 # Control structures
 # Ruby has the special 'unless' statement, which is syntactic sugar.
-# Unless (some condition) = If !(some condition).
+# Equivalent statements:
+#  unless some_condition
+#  if !some_condition
 
 if true
   'if statement'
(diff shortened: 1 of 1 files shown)


ruby.html.markdown
```diff
@@ -215,6 +218,17 @@ else
   'else, also optional'
 end
 
+unless true
+  'unless statement'
```
vendethiel: .... doesn't really explain what it does, though?

sdulal: I suppose it may be more descriptive if the statements were changed to "false statement' and 'true statement.'

vendethiel: yes, please do. (and add a "opposite of if" inside of the string, please)

ruby.html.markdown
```diff
@@ -206,6 +207,8 @@ new_hash.has_value?(3) #=> true
 # They share a lot of useful methods such as each, map, count, and more
 
 # Control structures
+# Ruby has the special 'unless' statement, which is syntactic sugar.
+# Unless (some condition) = If !(some condition).
```
sdulal: 'Unless' vs 'If' is given in this comment.

vendethiel: please inverse the "unless" condition still.

ruby.html.markdown
```diff
@@ -215,6 +218,17 @@ else
...
+  'true statement'
+else
+  'false statement, optional'
+end
+
+'if statement' if true
+'else statement' unless true
+
+(condition) ? 'true statement' : 'false statement'
+
```
onebree: I would change this entire part to `puts 'true statement'`, ect, as it gets the point across better the purpose of these conditionals. Furthermore, it is bad practice for a conditional to be `unless`/`else`, but rather `if true`/`else`.

Also, include a comment explaining how line 230 is uses the ternary operator.

ruby.html.markdown
onebree: I would rather put this into words, rather than with `=` signs:

```ruby
# Equivalent statements:
#  unless some_condition
#  if !some_condition
```

Issue #1333, opened 2015-10-04
csharp - The expression for the ternary operator is not very clear


guntbert:
currently it reads

    string isTrue = (true) ? "True" : "False";

The `(true)` is not really clear, a simple comparing expression explains it better.

When I submit a Pull Request, should  I do that for every affected language (fr, de, cn, tr, es) separately or all in one go?

thePsguy:
The Readme.md suggests separate pull requests for fixes in separate languages.

guntbert:
@thePsguy Good, I'll make just the one for English now, and after approval/discussion the others follow.

- Does that sound sensible to you? 
- It should be possible to have the commit in a feature-branch but still create the pull-request towards master, right?
- Would it be useful to have this issue mentioned in the PR?

<sub>I am asking because this is my first Contribution on GitHub</sub>

thePsguy:
-It is sensible, yes.
-Yes you can create changes in your branch and create pull requests.
-I myself am only 3 pull requests old, what is PR?

guntbert:
Pull Request :-)

thePsguy:
@guntbert, my bad! :D
If you mention the issue # in the description of the pull request, it gets mentioned here automatically! So it is useful in tracking issues.

PR #1342, opened 2015-10-04
[csharp/de,fr,tr,cn] Put demonstrative condition into ternary expression

guntbert:
It should be made clear that the part before the ternary operator is
indeed a condition, most often created as some comparison expression.

the same change for the English version has just been merged (see closed issue  #1333)


vendethiel:
ty! :+1:

guntbert:
@vendethiel Thx for the guidance :-), I learned several things :+1:

PR #1343, opened 2015-10-04
[python3/en]Add step parameter to range function

VeerpalB:
A description of the step parameter of the range function is added. An example of its use in a for loop is also given.


adambard:
Sounds good, thanks!

PR #1351, opened 2015-10-05
Update python.html.markdown

chashmeetsingh:


adambard:
This seems like an odd way to introduce `raw_input`; perhaps you could lead up to this? (Or, just demonstrate that `raw_input` returns the user's input as a string and leave the map up for them to figure out. This site attempts to cater to intermediate programmers, so we can leave the reader to piece together how to turn a string into a list of integers.

python.html.markdown
```diff
@@ -160,6 +160,10 @@ li = []
 # You can start with a prefilled list
 other_li = [4, 5, 6]
 
+# String slpit to list
```
levibostian: `slpit` => `split`

PR #1353, opened 2015-10-05
Reverse an array

chashmeetsingh:


adambard:
This method *does* exist, but could you include a bit to demonstrate that the array is reversed in-place and nothing is returned? Something like

```python
array = [1, 2, 3]
array.reverse()
print array  # Prints "[3, 2, 1]"
```

would be a bit better.

PR #1354, opened 2015-10-05
Reverse a String

chashmeetsingh:


adambard:
Haha, we all wish. `reverse` doesn't exist in python's standard library, I'm afraid you're stuck with [::-1].

PR #1355, opened 2015-10-05
[vim/en] Added vim.html.markdown

RadhikaG:
I've added a new tutorial for Vim as a tool. Please let me know if there's any changes I should make, or anything else I should add. Thanks!


vim.html.markdown
```diff
@@ -0,0 +1,231 @@
...
+editor designed for speed and increased productivity, and is ubiquitous in most 
+unix-based systems. It has numerous keybindings for speedy navigation to 
+specific points in the file, and for fast editing.
+
+## Basics of navigating Vim
+
+```
+    vim <filename>   # Open <filename> in vim
+    :w               # Save current file
+    :q!              # Quit vim without saving file
```
levibostian: Could we add a little more to the `!` character? It *forces* the command to execute, which in this case, is quit.

levibostian: Explain it a little more I mean. Instead of saying "quit vim without saving file" could explain that the `!` character forces the `:q` command to run which is quit leading to vim exiting without saving.

vim.html.markdown
```diff
@@ -0,0 +1,231 @@
...
+    :wq              # Save file and quit vim
```
levibostian: `:x` does the same thing. Want to add that along with this?

vim.html.markdown
```diff
@@ -0,0 +1,231 @@
...
+Vim is based on the concept on **modes**.
+
+Command Mode - vim starts up in this mode, used to navigate and write commands  
+Insert Mode  - used to make changes in your file  
+Visual Mode  - used to highlight text and do operations to them  
+Ex Mode      - used to drop down to the bottom with the ':' prompt to enter commands
+
+```
+    :                # Puts vim into ex mode
+    v                # Puts vim into visual mode
```
levibostian: Could you add `i` puts into insert mode?

levibostian:
Well done! I like this, thanks!

RadhikaG:
Thank you for the suggestions. I've added them all in. Can't believe I forgot to include 'i' for going into insert mode.

vendethiel:
:+1:

RadhikaG:
I'll be closing this pull request because I managed to screw up my branches in an odd way. I'll reopen the same pull request once my issue is sorted. Thanks.

PR #1367, opened 2015-10-06
[ruby-ecosystem/en] Ruby 1.9.x is also retired

theist:
Ruby 1.9.3 reached end of life in Feb 2015 as is stated in https://www.ruby-lang.org/en/news/2015/02/23/support-for-ruby-1-9-3-has-ended/ 

Is good to anyone to know that 2.x.x are the only rubies which get official security patches


vendethiel:
Right-o. Thanks!

PR #1371, opened 2015-10-06
[java/en] Added additional definitions and edited comments

zfergus2:
#### Added additional definitions and edited comments, Zachary Ferguson

Quick-list of changes:
* Included an example of printf
* Discussed final variable initialization
* Gave a floating point division example
* Discussed boolean operators
* Defined the abstract and final class and compared them.
* Added some clarifying remarks to comments.


levibostian:
Thanks!

PR #1373, opened 2015-10-07
Update ruby.html.markdown

kalebdavis:
Small change but useful, new programmers should know the difference between the two.


levibostian:
Thanks!

PR #1380, opened 2015-10-07
[objective-c/en] Added additional reading for both iOS and OS X for Objective-C

cwalk:
Added additional reading for both iOS and OS X for Objective-C from Apple Developer site. Added contribution line.


vendethiel:
Having links to the official docs can't be bad, can it? Thanks!

PR #1381, opened 2015-10-07
[swift/en] Added clearer description of Optionals and Unwrapping. 

cwalk:
Added clearer description of Optionals and Unwrapping. Minor typo changes as well. Also forgot to add myself as a contributor from swift-1 pull request.

### DIFF
--- a/swift.html.markdown
+++ b/swift.html.markdown
@@ -5,6 +5,7 @@ contributors:
   - ["Christopher Bess", "http://github.com/cbess"]
   - ["Joey Huang", "http://github.com/kamidox"]  
   - ["Anthony Nguyen", "http://github.com/anthonyn60"]
+  - ["Clayton Walker", "https://github.com/cwalk"]
 filename: learnswift.swift
 ---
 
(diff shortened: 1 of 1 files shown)


swift.html.markdown
```diff
@@ -74,11 +76,17 @@ if someOptionalString != nil {
...
-    
+
     let empty = someOptionalString?.isEmpty
 }
 someOptionalString = nil
 
+/*
+    To get the underlying type from an optional, you unwrap it using the 
+    force unwrap operator (!). Only use the unwrap operator if you're sure 
+    the underlying value isn't nil.
```
vendethiel: I don't think we should put such a statement. We're basically telling "always use `(!)`", which definitely isn't what we want to. the "if let" should look like a good solution as well

cwalk: True, good point. I was more trying to explain what ! was used for in swift, rather than implying it should be the only way to unwrap an optional. Maybe something along the lines of:

"Trying to use ! to access a non-existent optional value triggers a runtime error. Always make sure that an optional contains a non-nil value before using ! to force-unwrap its value."  

instead. Thoughts?

vendethiel: That seems good!

levibostian:
Thanks!

PR #1387, opened 2015-10-07
[brainfuck/nl] Fix dutch notation for large numbers

Iandenh:
The Dutch notation for large numbers is with a `.` instead of a `,`


vendethiel:
Thanks!

PR #1388, opened 2015-10-07
[ruby/sk] Slovak translation for Ruby

jkostolansky:


sk-sk/ruby.html.markdown
```diff
@@ -0,0 +1,553 @@
+---
+language: ruby
+filename: learnruby.rb
```
vendethiel: please add `-sk` suffix to the filename :).

PR #1392, opened 2015-10-07
[ruby/en] Add modulus, array.first, array.last, and each_with_index

ghalley:
I thought some simple demonstrations of these methods would be helpful, particularly when working with arrays.


ruby.html.markdown
```diff
@@ -258,6 +262,12 @@ hash.each do |key, value|
   puts "#{key} is #{value}"
 end
 
+# If you still need and index you can use "each_with_index" and define an index
+# variable
+array.each_with_index do |element, index|
+  pust "#{element} is number #{index} in the array"
```
levibostian: By `pust` do you mean `put`?

ghalley: oops.  That's exactly what I meant.  I can fix that.

levibostian:
Thanks!

PR #1397, opened 2015-10-08
[swift/en] Add generics operator in Swift

silvansky:
Just add a simple example of such operator.


vendethiel:
Sure -- thanks :).

PR #1399, opened 2015-10-08
example function added for call by reference

himanshu81494:
added swap function as an example for call by reference.
also added usage of "goto" in C.

### DIFF
--- a/c.html.markdown
+++ b/c.html.markdown
@@ -307,7 +307,25 @@ int main(void) {
     exit(-1);
     break;
   }
-
+  
+  // using "goto"
+  typedef enum { false, true } bool;
+  // for C don't have bool as data type :(
+  bool disaster = false;
+  int i, j;
+  for(i=0;i<100;++i)
+  for(j=0;j<100;++j)
+  {
+    if((i + j) >= 150)
+        disaster = true;
+    if(disaster)
+        goto error;
+  }
+  error :
+  printf("Error occured at i = %d & j = %d.\n", i, j);
+  // this will print out "Error occured at i = 52 & j = 99."
+  
+  
   ///////////////////////////////////////
   // Typecasting
   ///////////////////////////////////////
--- a/c.html.markdown
+++ b/c.html.markdown
@@ -307,8 +307,9 @@ int main(void) {
     exit(-1);
     break;
   }
-  
-  // using "goto"
+  /*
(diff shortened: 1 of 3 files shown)


c.html.markdown
```diff
@@ -472,7 +491,22 @@ char c[] = "This is a test.";
 str_reverse(c);
 printf("%s\n", c); // => ".tset a si sihT"
 */
-
+//as we can return return only one variable
```
levibostian: `return return` => `return`

c.html.markdown
```diff
@@ -472,7 +491,22 @@ char c[] = "This is a test.";
...
+//to change values of more than one variables we use call by reference
+void swapTwoNumbers(int *a, int *b)
+{
+int temp = *a;
```
levibostian: indent these, please to keep with style of rest of doc.

c.html.markdown
```diff
@@ -306,7 +307,25 @@ int main(void) {
...
     break;
   }
-
+  
+  // using "goto"
+  typedef enum { false, true } bool;
+  // for C don't have bool as data type :(
+  bool disaster = false;
+  int i, j;
+  for(i=0;i<100;++i)
```
levibostian: Two for loops stacked on top of each other, this is possible?

c.html.markdown
himanshu81494: sorry i did not  understand what "return return => return " means.

c.html.markdown
himanshu81494: https://ideone.com/GuPhd6

please correct me if I am somewhere wrong. This is my first ever pull request on git !

c.html.markdown
vendethiel: you have two times the "return" word

c.html.markdown
himanshu81494: OK, I did it.

c.html.markdown
himanshu81494: got it 
I have corrected it.

c.html.markdown
levibostian: No, you are not mistaken. I don't know why I thought it was not possible, it is just nested loops not that weird. I had just never seen it done like this, looked weird or something.

levibostian:
Thanks!

PR #1404, opened 2015-10-08
[perl6] fix the output of ff example

tecknicaltom:
the output for this example of the ff operator is wrong. This PR fixes it


levibostian:
Thanks!

PR #1409, opened 2015-10-08
[en/git] Add git-clean info

symm:
How to clean up untracked files in Git

### DIFF
--- a/git.html.markdown
+++ b/git.html.markdown
@@ -220,6 +220,25 @@ $ git checkout branchName
 $ git checkout -b newBranch
 ```
 
+### clean (caution)
+
+Remove files not under version control
+
+```bash
+# Given we have two untracked files
+$ git status -s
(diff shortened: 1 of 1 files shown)


git.html.markdown
```diff
@@ -220,6 +220,25 @@ $ git checkout branchName
...
+Remove files not under version control
+
+```bash
+# Given we have two untracked files
+$ git status -s
+A  game.c
+?? tmp-file
+?? untracked
+# When we force clean
+$ git clean -f
```
levibostian: Want to add note you can also use `-d` to clean directories, not just files?

onebree:
I think this is important to note, from the manpages, that `-f` may not be necessary

```
-f, --force
    If the Git configuration variable clean.requireForce is not set to false,
    git clean will refuse to run unless given -f or -n.
```

PR #1410, opened 2015-10-08
[matlab/en] Added simple class example to Matlab

### LINKED ISSUE #1038: [MATLAB] Add class example
An example of how to define and instantiate classes for MATLAB would be helpful.
The [MATLAB docs](http://uk.mathworks.com/help/matlab/matlab_oop/matlab-vs-other-oo-languages.html#brqzfut-8) have:
```matlab
classdef SimpleClass
   properties
      Color
   end
   methods
      function obj = SimpleClass(c)
         if nargin > 0
            obj.Color = c;
         end
      end
   end 
end
```
If I were more familiar with MATLAB I would add this myself, unfortunately I am not, so am unsure what best practices are.

voltnor:
Added a simple example to explain Matlab classes. Classes in Matlab almost seem like an afterthought by Mathworks, so I thought they'd be appropriately at home at the end of the document. 

Closes  #1038

### DIFF
--- a/matlab.html.markdown
+++ b/matlab.html.markdown
@@ -3,6 +3,7 @@ language: Matlab
 contributors:
     - ["mendozao", "http://github.com/mendozao"]
     - ["jamesscottbrown", "http://jamesscottbrown.com"]
+    - ["Colton Kohnke", "http://github.com/voltnor"]
 
 ---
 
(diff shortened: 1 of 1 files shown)


levibostian:
Thanks!

Issue #1119, opened 2015-05-29
[Request] add LaTeX


sangimed:

ramananbalakrishnan:
duplicate of #815 , might as well close both issues, since PRs #1412 and #1821 solve them.

Oire:
Fixed.

PR #1412, opened 2015-10-08
[latex/en] Initial latex whirlwind showcase revisited

voltnor:
Revisited the initial latex quick reference and reorganised its structure. Instead of multiple comments and commands, created a compile-able latex document to show off the bare-bones basic structure of latex documents and a few built-in features. 

Includes pull request #1055 
Closes issue #1119

### DIFF
--- a/latex.html.markdown
+++ b/latex.html.markdown
@@ -2,6 +2,7 @@
 language: latex
 contributors:
     - ["Chaitanya Krishna Ande", "http://icymist.github.io"]
+    - ["Colton Kohnke", "http://github.com/voltnor"]
 filename: learn-latex.tex
 ---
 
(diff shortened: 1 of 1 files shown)


latex.html.markdown
```diff
@@ -0,0 +1,154 @@
...
+containing a lot of mathematics. Getting a good document is very easy, but
+getting it to behave exactly the way you want can be a bit hairy.
+
+
+```latex
+
+% All comment lines start with %
+% There are no multi-line comments
+
+% LaTeX is NOT a ``What You See Is What You Get'' word processing software like
```
levibostian: You use a mix of `` characters and ''. Please choose one or other.

latex.html.markdown
```diff
@@ -0,0 +1,154 @@
...
+
+\section{Some Text notes}
+LaTeX is generally pretty good about placing text where it should go. If 
+a line \\ needs \\ to \\ break \\ you add \textbackslash\textbackslash to 
+the text. In case you haven't noticed the \textbackslash is the character
+the tells the LaTeX compiler it should pay attention to what's next.
+
+\section{Math}
+
+One of the primary uses for LaTeX is to produce academic article or
```
levibostian: `produce an academic`. Add `an` please.

levibostian:
Thanks for this. 

I noticed that some of the lines of the document explaining various parts of Latex has the lines start with `%` indicating a comment while other lines don't. So did you do that on purpose or do you need to go through all lines of this file and add `%` characters to comment them all out?

voltnor:
What I was trying to do was show off the bare bones basic features of latex (figures, tables, captions, sections, etc) in the compiled pdf document that gets generated. Then if the user wants to use one of those features, they could go to the source code and see how it's done, with comments. In this sense, comments about the features would belong in the compiled document and comments about how to implement the feature would belong in the source code. 

Latex is strange to learn in this project's format as there are benefits from seeing the source code and the compiled document side by side. This is why I chose to make the latex code above actually compile instead of being code snippets. I'd be interested in other opinions on presentation though.

I'll take another look at which comments are where, and there is a pull request on my branch that addresses your other comments and cleans some things up more. I'll hopefully have an updated branch to pull later tonight.

levibostian:
Thanks!

PR #1413, opened 2015-10-08
[en/git] Add Git Immersion tutorial as a resource

hfenton:
Added [Git Immersion](http://gitimmersion.com/) as a resource


levibostian:
Thanks!

PR #1416, opened 2015-10-09
Mention of trailing commas in JSON and that they should be avoided

### LINKED ISSUE #854: Commas after last element
It should be noted that JSON versions vary about using trailing commas or not.
General rule of thumb is not to use trailing commas for comaptibility reason, although this may complicate some code.
Also people started to add commas before element to avoid commas on the end.

More discussion:
http://stackoverflow.com/questions/201782/can-you-use-a-trailing-comma-in-a-json-object

AVGP:
fixes #854


adambard:
Yes! Thank you.

PR #1423, opened 2015-10-09
[objective-c/en-fr-ru-tr-vn] Enable syntax highlighting

AndyBrown91:
The syntax highlighting was using the wrong code


levibostian:
Thanks!

PR #1425, opened 2015-10-09
[C/en] Accessing command line arguments in main

AndyBrown91:
Also added a note about the convention of constant names being all caps


levibostian:
Thanks!

PR #1426, opened 2015-10-09
[xml/en] Grammar fixes

AndyBrown91:
Children makes more sense grammatically and is found in a lot of tutorials online, so makes sense to include it here


vendethiel:
Ty! :+1:

PR #1427, opened 2015-10-09
[c/en] Added a section for header files.

zfergus2:
## [c/en] Added a section for header files.

Quick-list of changes:
* Added a section for header files.
* Included a discussion of what belongs in a header file and what does
not.


c.html.markdown
```diff
@@ -630,6 +630,54 @@ typedef void (*my_fnp_type)(char *);
...
+Header files are an important part of c as they allow for the connection of c 
+source files and can simplify code and definitions by seperating them into 
+seperate files.
+
+Header files are syntaxtically similar to c source files but reside in ".h" 
+files. They can be included in your c source file by using the precompiler 
+command #include "example.h", given that example.h exists in the same directory 
+as the c file.
+
+```c
```
levibostian:
```c is already declared in the file so it is not needed here.

levibostian: I see line 631 ends the syntax highlighting. Would be best if you please remove the ``` on line 631 and this line here.

zfergus2: This was intended to be the start of a seperate section on header files, ".h" files. Since header files follow the same syntax of c files "```c" was used to mark the begining of the header's code.

zfergus2: I went ahead and changed the ```c to a ```h for the start of the header file's code section.

levibostian: That wont work actually. It must be 'c' as that is the syntax highlighting. 

Go ahead and remove the ``` from line 631 and this line here we are talking about and we will get it merged in.

zfergus2:
I went ahead and removed the lines in question. I still believe, however, that since c files and h, header, files are separate files when programming, there should be separate code sections in the markdown.

levibostian:
Oh, that is why you made them separate...I thought you did that by mistake...

First off, can you please merge master to remove the conflict? I understand the reason behind having them different sections, but I also don't believe header files should be a different section completely in the document, there should be a header file for all of the c files in the doc.

zfergus2:
Merged with main

vendethiel:
Thanks!

PR #1436, opened 2015-10-10
Updated Documentation

chashmeetsingh:


json.html.markdown
```diff
@@ -37,7 +37,9 @@ For the purposes of this, however, everything is going to be 100% valid JSON. Lu
       "comment": "These things can be nested, very useful."
     }
   },
-
+  
+  "string" : ["json array containing json objects"],
```
levibostian: could you change the key from "string" to "array"? It is confusing as reading this I think that `["json ... objects"]` is a string but that is not the case.

PR #1444, opened 2015-10-10
Added document about Qt framework

vortexxx192:
Don't know in which format code to download should be added so I put it in a separate file.
Please share your thoughts. Thank you.


adambard:
Thanks for the contrib!

The code file is generated from all the code blocks in the markdown file concatenated together. Do you think that will cause problems with your file?

vortexxx192:
Oh, I have just made the work that is automated -- my code file is an exact concatenation of all code blocks from markdown file.

vortexxx192:
So should I now make one more commit with qt.cpp removed?

vendethiel:
please embed the `qt.cpp` file :).

vendethiel:
@vortexxx192 ? :)

vortexxx192:
@vendethiel Oh sorry, missed your previous comment.

Well now I have made one more commit, is everything OK now? To be fair, I just don't get what "please embed the qt.cpp file" exactly means.

vendethiel:
LGTM! Thank you.

PR #1316, opened 2015-10-03
Add pt-pt translation for Scala tutorial

JD557:
This PR adds a Portuguese translation of the Scala tutorial.

I tested the code and it seems to be working as expected. This is my first pull request for this project though, so I might have missed something from the guidelines (eg. do I need to include the code in a separate `learn-pt.scala` file somewhere, or is it automagically generated?).

PR #1446, opened 2015-10-10
Fix compile errors of the English and French Scala tutorials

JD557:
The Scala tutorials had some small bugs, namely:

* Attempting to use an immutable variable `x` to control a loop;
* Uncommented code that would not compile, as it's not valid scala;
* Undefined lists and case classes;
* Two filenames;

This errors are also fixed on the pt-pt version (PR #1316).


vendethiel:
Thanks!

PR #1451, opened 2015-10-10
Add example of post-condition for loop.

bentglasstube:
Post condition ifs are mentioned, but not post condition for loops.  Post condition for loops are pretty widely used in my experience so I felt they should be listed.


vendethiel:
Thanks!

Issue #1452, opened 2015-10-10
[bash/de] The translation is out of date


guntbert:
bash.html.markdown contains sections that are missing in the German version
- a lot about variables
- conditional execution
- redirection
- loops
- functions
- help
- man
- info

guntbert:
Pull request is on the way - will be Work in Progress for a few days

guntbert:
What is needed to get this on its way? I *think* my work should be done.

vendethiel:
Seems fine from here. thanks

=> #1703 1453

PR #1449, opened 2015-10-10
[git/de] Improve the translation

guntbert:
several characters were wrong, additionally German spelling/grammar is
now correct (German is my primary language).

PR #1453, opened 2015-10-10
[bash/de] #1452 Several sections need to be translated

guntbert:
The first two sections
- variables
- conditionals

rest will follow in the next days


guntbert:
I just noticed - commit 21c1685 doesn't belong here (part of #1449) , its the same mistake I made last time
but reverting it won't help either - so I'll do the same like with 1342 :-)

guntbert:
My work seems to be done finally - please review

PR #1454, opened 2015-10-10
[matlab/it] Add matlab italian translation

fontealpina:


it-it/matlab-it.html.markdown
```diff
@@ -0,0 +1,525 @@
+---
+language: Matlab
```
vendethiel: please add a filename field

PR #1455, opened 2015-10-10
[css/en] corrected typo in css doc

vaichidrewar:
corrected typo in css doc

### DIFF
--- a/css.html.markdown
+++ b/css.html.markdown
@@ -108,7 +108,7 @@ div.some-parent.class-name {}
    based on its page behaviour (rather than page structure) */
 
 /* for example for when an element is hovered */
-selector:hover {}
+selected:hover {}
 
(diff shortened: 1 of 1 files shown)


vendethiel:
How is that a typo?

vaichidrewar:
May be I should not have marked it as a typo. 
Usage of all the pseudo classes have the same selector element mentioned before colon.
So in all the four examples shouldn't the name before column be consistent? Either user 'selector' in all places or use 'selected'. Am I missing something. 

selector:hover { }

/* or a link has been visited */
selector:visited { }

/* or hasn't been visited */
selected:link { }

/* or an element in focus */
selected:focus { }

vendethiel:
That's right. But maybe we should change it to `selector` everywhere, then? Not to be confused with `:selected`?

PR #1460, opened 2015-10-11
Added C# Expression Body Definition example

csuich2:


csuich2:
When I started learning C#, I found this to be a really cool & helpful trick/feature.

PR #1461, opened 2015-10-11
Added detail on C# single inheritance

csuich2:
Clarified that the base class name must be the first thing listed


levibostian:
Thanks!

PR #1469, opened 2015-10-11
[C++/en] Small inheritance clarification

Alters:
Clarified that private members are inherited but are not directly
accessible


levibostian:
Thanks!

PR #1471, opened 2015-10-11
[python/en] Added instructions on how to perform input operations

ankitaggarwal011:
Added instructions on how to perform input operations in Python 2.7. 
I believe this is an important instruction currently missing from this document. Please review it, @vendethiel.

### DIFF
--- a/python.html.markdown
+++ b/python.html.markdown
@@ -4,7 +4,6 @@ contributors:
     - ["Louie Dinh", "http://ldinh.ca"]
     - ["Amin Bandali", "http://aminbandali.com"]
     - ["Andre Polykanine", "https://github.com/Oire"]
-    - ["Ankit Aggarwal", "http://ankitaggarwal.me"]
 filename: learnpython.py
 ---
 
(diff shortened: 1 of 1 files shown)


python.html.markdown
```diff
@@ -144,6 +145,10 @@ bool("")  # => False
 # Python has a print statement
 print "I'm Python. Nice to meet you!"
 
+# Simple way to get input data from console
+input_string_var = raw_input("Enter some data: ") # Data is stored as a string
+input_number_var = input("Enter a number: ") # Data is stored as a number
```
levibostian: Actually, data is not stored as a number all the time. If I enter "learn x in y" for the input prompt, it will be a string, not an int. So it *can* be an int but also *can* be something else.

ankitaggarwal011: @levibostian , Hi, my mistake, thank you for correcting. input() treats the input data as python code and tries to execute it. Yes, It evaluates numbers as well as python code given as input data.
Normally, it's use is not recommended and it's been deprecated in Python 3. Should input() method be actually included in this documentation? What do you think? Please let me know, I will commit the changes. Thanks.

ankitaggarwal011:
@levibostian, I committed the changes. Please review these and let me know. Thanks.

levibostian:
That works for me. Thanks!

PR #1477, opened 2015-10-12
Fixed the custom indexer example (setter return type is void)

jdpearce:
Just a small fix to the setter example and I altered the word "yields" to "returns" due to "yield" being a keyword which might confuse.

There are now a ton of warnings, but after installing entity framework the code as it stands does compile! :)


levibostian:
Thanks!

PR #1478, opened 2015-10-12
Range is not a generator

aayushranaut:
Remove the line describing the range is a generator.
Also removed useless spaces at the end of lines.

adambard#1323


levibostian:
Thank you for taking care of this, appreciate it!

verhovsky:
`range` is a generator though.

doolio:
I don't believe it is. It is an iterable object that you can get an iterator from. Can you point me to official documentation that states that it is a generator?

PR #1484, opened 2015-10-12
Added BigInteger and BigDecimal explanations to Java tutorial

RobRoseKnows:
Brief explanations of what BigInt and BigDecimal are for and what they consist of. Also added the ranges for float and double according to Java documentation.


levibostian:
Thanks!

PR #1485, opened 2015-10-12
git-fr.html.markdown

MrBricodage:
Translation of git.html.markdown into french, using new file git-fr.html.markdown


fr-fr/git-fr.html.markdown
```diff
@@ -0,0 +1,509 @@
+---
+category: tool
+tool: git
+filename: LearnGit.txt
```
vendethiel: please add a `-fr` filename suffix

PR #1493, opened 2015-10-13
[python/en] Add resources to the "Free Online" section

evuez:
These have been valuable resources to me, but I'm not sure whether it fits this section or not.

The PEP8 should definitely be there imho, but we might want to add another section for this kind of resources.


levibostian:
Awesome Python FTW! Thanks

PR #1497, opened 2015-10-13
[json/en] Fixed incorrect information about commas in JSON doc

mwozniczak:
Original version indicated that commas are placed before _values_ in JSON, which is incorrect.


levibostian:
Thanks!

PR #1503, opened 2015-10-14
Update Python 2 vs Python 3

Carreau:
Add fact that Python 2 is reaching end of life. 
And note that you can use `__future__` to add Python 2 and 3 compatible code.


levibostian:
Thanks!

Carreau:
Thanks !

PR #1512, opened 2015-10-14
[c/en] clarify common tripping point of newbies

awalGarg:
```
int foo () {
  printf("bar\n");
  int x; // this is not valid in C89+
}
```


levibostian:
Thanks!

PR #1518, opened 2015-10-14
[markdown/en] remove unreadable backslashes

awalGarg:
the recommended way to markdown inside markdown is whitespace characters :-)


awalGarg:
welp, this didn't happen right. looking into it, please don't merge

awalGarg:
weird. Can someone diff between the current version in this branch and https://github.com/awalGarg/learnxinyminutes-docs/blob/patch-4/markdown.html.markdown? the weird `\\` in the indented ruby block are fixed in my repo but for some reason github web gui doesn't show the changed diff here.

markdown.html.markdown
```diff
@@ -153,12 +153,11 @@ inside your code -->
 John didn't even know what the `go_to()` function did!
 
 <!-- In Github Flavored Markdown, you can use a special syntax for code -->
-
-\`\`\`ruby <!-- except remove those backslashes when you do this, just ```ruby ! -->
-def foobar
-    puts "Hello world!"
-end
-\`\`\` <!-- here too, no backslashes, just ``` -->
+    ```ruby
```
levibostian: What happens if you remove the whitespace? Will it break the markdown or something? I think the whitespace will confuse users.

awalGarg: @levibostian If you remove the whitespace, it marks an end to the code block started at the start of the block marked "markdown" which is why I believe "daturkel" introduced the hacky backslashes in this commit: https://github.com/awalGarg/learnxinyminutes-docs/commit/55ded27c4f755c36e1ec09d681e76a7df418e6a9#diff-5744dd7bd81f21547f768566a7e0bc34R135. But markdown (http://spec.commonmark.org/0.22/#indented-code-blocks) allows us to indent code with spaces like this so we can give up the backslashes for a cleaner solution :)

levibostian: If the indenting is necessary, please make a note of that in the doc. It currently does not state *why* there is an indent.

awalGarg: @levibostian fixed by https://github.com/awalGarg/learnxinyminutes-docs/commit/3c28b43c731f4f99662fbbb0413a03126b1ea150

markdown.html.markdown
```diff
@@ -153,12 +153,13 @@ inside your code -->
...
 
 <!-- In Github Flavored Markdown, you can use a special syntax for code -->
-
-\`\`\`ruby <!-- except remove those backslashes when you do this, just ```ruby ! -->
-def foobar
-    puts "Hello world!"
-end
-\`\`\` <!-- here too, no backslashes, just ``` -->
+(the extra indentation is only to render the fenced block as-is here.
+omit it in your markdown) -->
```
onebree: Is the `-->` supposed to be here? IMO, remove the statement from parentheses, as it is important and yet another markdown tip for the cheat sheet.

awalGarg: @onebree Right, dunno how I missed that. fixed

onebree:
@awalGarg I tried this myself. I am not sure if it is the way Github uses their markdown, but I had an issue. First, if you use spaces, Markdown will include those, and the code will be unnecessarily indented:

```markdown
    ```ruby
    puts "Hello, #{planet}!"
    ```
```

Now, if I were to remove the extra spaces (otherwise, the same code), this happens. The second Markdown block appears, and is rendered as not being closed (even though it looks like it is). (I closed it so this comment could render correctly.)

```markdown
```ruby
puts "Hello, #{planet}!"
```
```
```

You can trick the parser into closing the ruby block with:

    ```-

Or any non-space character.

PR #1521, opened 2015-10-14
[xml/en] Superfluous spaces removed in some tags in 'XML'

mwozniczak:
In latter examples, some <title> tags had spaces near the end of the opening tag (possibly because later examples omit the 'lang' attribute of it)


vendethiel:
Thanks

PR #1524, opened 2015-10-14
[latex/es] Spanish translation for the LaTeX article

csuarez:
This is the first Spanish translation for the LaTeX tutorial. I've tried to keep the translation natural for the Spanish reader, so I've took some licenses during the translation, but without changing the meaning of the text in any case.


vendethiel:
sorry, it seems #1958 was merged first.

we received a lot of pull requests in october and are only now catching up... sorry for the delay.

PR #1525, opened 2015-10-14
[latex/en] Fix some typos in LaTeX article

csuarez:
I've fixed some little typos in this article. Maybe the most important is to reference everywhere LaTeX as "LaTeX" (neither "latex" nor "Latex"), like as the beginning of the article. It's a little bit weird to read it referenced in three different ways.


csuarez:
I've added another commit, I replaced too many "latex" words :stuck_out_tongue:

levibostian:
Thanks!

PR #1526, opened 2015-10-15
[java/en] Enum Type

raphaelbn:
Short overview about enum type.

### DIFF
--- a/java.html.markdown
+++ b/java.html.markdown
@@ -7,7 +7,7 @@ contributors:
     - ["Simon Morgan", "http://sjm.io/"]
     - ["Zachary Ferguson", "http://github.com/zfergus2"]
     - ["Cameron Schermerhorn", "http://github.com/cschermerhorn"]
-    - ["Raphael Nascimento", "http://github.com/raphaelbn"]
+    - ["Rachel Stiyer", "https://github.com/rstiyer"]
 filename: LearnJava.java
 ---
 
@@ -138,7 +138,7 @@ public class LearnJava {
         //
         // BigDecimal allows the programmer complete control over decimal
         // rounding. It is recommended to use BigDecimal with currency values
-        // and where exact decimal percision is required.
+        // and where exact decimal precision is required.
         //
         // BigDecimal can be initialized with an int, long, double or String
         // or by initializing the unscaled value (BigInteger) and scale (int).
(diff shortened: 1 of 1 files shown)


java.html.markdown
```diff
@@ -670,6 +671,68 @@ public abstract class Mammal()
...
+                        
+            default:
+                System.out.println("Midweek days are so-so.");
+                break;
+        }
+    }
+    
+    public static void main(String[] args) {
+        EnumTest firstDay = new EnumTest(Day.MONDAY);
+        firstDay.tellItLikeItIs();
```
levibostian: Instead of having the outputs below, can you put them here inline? 
`firstDay.tellItLikeItIs(); // => Mondays are bad.`

raphaelbn:
Hi @levibostian,

Sure, I've already made the commit.

:+1:

raphaelbn:
Is it ok now, @levibostian ?

levibostian:
Thanks! Now can you merge in master please?

raphaelbn:
Hi @levibostian , I guess I don't have permission to merge..

raphaelbn:
@levibostian, I updated the code and it's not in conflit anymore. 
Can you merge, please?

geoffliu:
Wait wait, one comment: The example given is not really in Java style. Enums in Java can have methods on them, and in the example you gave, the `tellItLikeItIs` method is better in the enum class itself, so you can do 

    Days.MONDAY.tellItLikeItIs()

I would rather show that as an example, and if you really want to show how enums are used in other classes, make up a use that is not intrinsic to the enum class (and therefore belongs in the enum class). I'm thinking something like a separate `Person` class that has `isWorkingOn(Day day)`

raphaelbn:
It's only a little example how basic enum works. I know enum can have methods and be better than I've demonstrated.
I said:
"
// Enum types are much more powerful than we show above. 
// The enum body can include methods and other fields.
// You can se more at https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html
"

And this example is in the oracle oficial doc:
https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html

raphaelbn:
@geoffliu

geoffliu:
Still rather have a slightly different example, but I won't press the point. :smile: 

Could you bring it up to date with master, and I'll merge? Thanks!

raphaelbn:
you're welcome @geoffliu :smile:

It's already done.

Thanks :+1:

PR #1527, opened 2015-10-15
[d/en] Next set of improvements

gkotian:
Following on from #1495.

### DIFF
--- a/d.html.markdown
+++ b/d.html.markdown
@@ -74,16 +74,18 @@ are passed to functions by value (i.e. copied) and classes are passed by referen
 we can use templates to parameterize all of these on both types and values!
 
 ```c
-// Here, T is a type parameter. Think <T> from C++/C#/Java
+// Here, 'T' is a type parameter. Think '<T>' from C++/C#/Java.
 struct LinkedList(T) {
     T data = null;
-    LinkedList!(T)* next; // The ! is used to instaniate a parameterized type. Again, think <T>
+
+    // Use '!' to instantiate a parameterized type. Again, think '<T>'.
+    LinkedList!(T)* next;
 }
 
 class BinTree(T) {
     T data = null;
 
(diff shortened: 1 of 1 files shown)


d.html.markdown
```diff
@@ -95,16 +97,14 @@ enum Day {
     Wednesday,
     Thursday,
     Friday,
-    Saturday,
```
vendethiel: why?

gautam-kotian-sociomantic: Well... I guess most compilers wouldn't complain about the trailing comma. But since it is of no use, I feel it is better not to have it at all. This is particularly true in case of enums like in this example (days of the week) where no new fields are going to be added. If we were dealing with an enum that one could expect to grow, then I may be ok with that trailing comma because a new field added to the end wouldn't cause the previous line to change (thus marginally simplifying a review). What do you think?

gkotian: Well... I guess most compilers wouldn't complain about the trailing comma. But since it is of no use, I feel it is better not to have it at all. This is particularly true in case of enums like in this example (days of the week) where no new fields are going to be added. If we were dealing with an enum that one could expect to grow, then I may be ok with that trailing comma because a new field added to the end wouldn't cause the previous line to change (thus marginally simplifying a review). What do you think?

rightfold: Trailing commas make diffs nicer, make merges easier, and is allowed explicitly by D and any compiler that doesn't support them is in error.

gautam-kotian-sociomantic: Ok, fair enough. Will remove the offending commit.

gkotian: Ok, fair enough. Will remove the offending commit.

gkotian:
Repushed after removing the commit that removed the trailing comma.

vendethiel:
Thanks! :cake:

PR #1528, opened 2015-10-15
[JSON/cs] Czech translation for JSON

scheras:


cs-cz/json.html.markdown
```diff
@@ -0,0 +1,70 @@
+---
+language: json
+filename: learnjson.json
```
vendethiel: please add a `-cz` filename suffix.

vendethiel:
Seems like the document was translated in the meantime. Thanks still!

PR #1540, opened 2015-10-15
[C/en] More info on dynamically allocated arrays

bbatliner:
Dynamically allocated arrays and pointers can be confusing. I added some examples and clarifications.


levibostian:
Thanks!

PR #1544, opened 2015-10-15
[php/en] Adds late static binding

Glutnix:
There was no reference to late static binding in the document, so added a working example. Let me know if this needs work.


levibostian:
Thanks!

PR #1547, opened 2015-10-16
Update Julia variable names

### LINKED ISSUE #516: [julia]Variable names start with a letter
I'm not Julia user so I don't want to change anything but I have tried creating variables and version 0.2 lets you start variable with underscore.

bharrisau:
Julia supports underscores at the beginning of a variable name now. It also supports a subset of Unicode code points for the first character, and additional code points for subsequent characters.

A full explanation of all available code points was excluded as it is quite complicated and doesn't add much value in the X in Y context.

closes #516

### DIFF
--- a/julia.html.markdown
+++ b/julia.html.markdown
@@ -117,11 +117,11 @@ catch e
     println(e)
 end
 
-# Variable names start with a letter.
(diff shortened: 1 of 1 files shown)


levibostian:
Thanks!

PR #1548, opened 2015-10-16
Update r.html.markdown

### LINKED ISSUE #427: [r/en] Function in Apply Section not defined
In the apply() family of functions, you might want to define MyFunc before using it.

Also, right at the start: in Windows it's CTRL-RETURN (or CTRL-ENTER); there is no CMD button (that's for Macs).

Thanks
Steve

bharrisau:
Correct key used to execute a line on Windows.

closes #427


levibostian:
Thanks!

PR #1549, opened 2015-10-16
Fix misspelled word Boolean

jaminscript:
Boolean is named after George Boole, and is supposed to start with a
capital. I only updated English comments as I am not sure what the
impact of this has on other spoken languages.

http://english.stackexchange.com/questions/4481/should-the-word-boolean-be-capitalized

### DIFF
--- a/clojure.html.markdown
+++ b/clojure.html.markdown
@@ -52,7 +52,7 @@ and often automatically.
 ; Types
 ;;;;;;;;;;;;;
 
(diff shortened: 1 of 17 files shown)


levibostian:
I am going to close this for now. It is going to be tough to merge as we can see, merge conflicts are super easy to happen as so many files are edited. 

So if you want this merged in, please make multiple PRs and so we can actually merge it without conflicts. 

Thanks for your contribution!

PR #1555, opened 2015-10-16
[bash/ro] Changed some old stuff and added diacritics

ianiD:
Tweaked around with the old translations


vendethiel:
Thanks & sorry for the delay!

PR #1557, opened 2015-10-16
[make/en] add .SILENT rule to make

awalGarg:


awalGarg:
err, the `@` symbol is already covered so I guess this would better not be added. My bad :(

PR #1560, opened 2015-10-16
[python/en] Adding documentation on ternary operator

ankitaggarwal011:
Adding documentation on C's '?:' ternary operator equivalent in Python. 
Since, a lot of people from C/C++/Java might use this documentation, they would look for a ternary operator equivalent in Python. 
This was not mentioned explicitly, even though ternary operator is used in the documentation.

What do you think? @levibostian, @adambard. Please let me know. Thanks.

### DIFF
--- a/python.html.markdown
+++ b/python.html.markdown
@@ -165,6 +165,7 @@ some_var  # => 5
 some_other_var  # Raises a name error
 
 # if can be used as an expression
+# Equivalent of C's '?:' ternary operator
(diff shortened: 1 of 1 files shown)


levibostian:
Thanks!

levibostian:
Can you add this change to python3 now?

ankitaggarwal011:
Yes, sure. I will do that today.

PR #1564, opened 2015-10-16
[latex/en] Fixed markdown syntax, now this file is markdown friendly

TonnyORG:
I just make this document friendly for markdown due if you go to http://learnxinyminutes.com/docs/latex/ you will see that it's a plain text and there's no code to show up.


levibostian:
Hmmm...I am not sure if this change is needed or not. For the docs on the site, we like to keep everything as 1 file that you are able to download and execute, not just read it from the website. So that is why we have one giant chunk of code with comments explaining each piece. 

So I am going to close this PR for now as the document right now is one giant code chunk you are able to download as a latex source code file.

PR #1574, opened 2015-10-16
[hy] Updating, elaborating and cleaning some rust

Zirak:
Added some more recent hy changes (class syntax, `cut` instead of slice), as well as some missing operators (`->` and `->>`) and function declarations (using `&rest` and `&kwargs` along with `apply`).

Also changed the code block syntax to hy since github recognises it.

### DIFF
--- a/hy.html.markdown
+++ b/hy.html.markdown
@@ -12,7 +12,7 @@ hy to call native python code or python to call native hy code as well
 
 This tutorial works for hy ≥ 0.9.12, with some corrections for hy 0.11.
 
-```hy
+```clojure
 ;; this gives an gentle introduction to hy for a quick trial head to
 ;; http://try-hy.appspot.com
 ;;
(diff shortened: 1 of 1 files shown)


hy.html.markdown
```diff
@@ -3,15 +3,16 @@ language: hy
...
 
 Hy is a lisp dialect built on top of python. This is achieved by
 converting hy code to python's abstract syntax tree (ast). This allows
 hy to call native python code or python to call native hy code as well
 
-This tutorial works for hy ≥ 0.9.12
+This tutorial works for hy ≥ 0.9.12, with some corrections for hy 0.11.
 
-```clojure
+```hy
```
adambard: I'm afraid rouge doesn't support hy highlighting yet, clojure is the closest you're going to get.

adambard: To be clear, this will cause an error and you'll need to change it back :P

vendethiel:
Thanks!

PR #1576, opened 2015-10-16
Added Templates to C++ Tutorial

brunoalano:


cmake.html.markdown
```diff
@@ -0,0 +1,176 @@
+---
```
vendethiel: why is cmake here

brunoalano: Sorry, issue when commiting

vendethiel:
Thanks & sorry for the delay

PR #1577, opened 2015-10-17
Update PHP with Additional Information

AkshayKalose:


php.html.markdown
```diff
@@ -12,7 +12,7 @@ This document describes PHP 5+.
 <?php // PHP code must be enclosed with <?php tags
 
 // If your php file only contains PHP code, it is best practice
-// to omit the php closing tag.
+// to omit the php closing tag which will Pprevent accidental output.
```
adambard: Pprevent -> prevent

And probably, "which will" -> "to"

adambard:
Thanks!

PR #1578, opened 2015-10-17
Add Getting Parameters From Functions in PHP

AkshayKalose:


adambard:
Thanks! Do you think you could change the inventing to 4 spaces to match the rest of the doc?

AkshayKalose:
Fixed! Looks like the whole functions section was off style.

adambard:
Ah, so it was. Thanks for taking care of it!

PR #1586, opened 2015-10-17
Adding some documentation about zfs

sarlalian:
ZFS is a great filesystem, that many developers and sysadmins may come across in their careers. Thought people might like a basic guide.


adambard:
Neat, thanks!

PR #1587, opened 2015-10-17
[YAML/nl] Translated YAML to dutch

HetIsNiels:
Translated YAML document to dutch


adambard:
Thanks! I think both of the dutch programmers who aren't fluent in English will appreciate this ;)

PR #1589, opened 2015-10-17
[C++/en] Added in Enum information

Alters:
Added in some information about enums in C++ including some different
use cases


adambard:
Thanks!

PR #1594, opened 2015-10-17
Adds german translation for LateX

phxql:
Hi,

this adds the german translation for the LateX tutorial.

Greetings,
Moritz


adambard:
Danke!

PR #1597, opened 2015-10-17
[ruby/en] Adds more string operations to ruby doc

pzula:
Adding more string operations that are sometimes seen in open source projects.


ruby.html.markdown
```diff
@@ -107,6 +108,12 @@ placeholder = 'use string interpolation'
 'hello ' + 3 #=> TypeError: can't convert Fixnum into String
 'hello ' + 3.to_s #=> "hello 3"
 
+# Combine strings and operators
+'hello ' * 3 #=> "hello hello hello"
```
adambard: Doesn't this return `"hello hello hello "` (with a trailing space)?

pzula: Whoops! Yes it does :) Fixing it now

adambard:
Thanks!

PR #1608, opened 2015-10-18
[javascript/en] Adding modulo division

venegu:
This pull request simply adds examples of modulo division and their expected outcomes to the JavaScript article. :panda_face:


adambard:
Thanks!

PR #1621, opened 2015-10-18
[JSON/en] JSON.parse() method

raphaelbn:


json.html.markdown
```diff
@@ -64,4 +64,30 @@ Drawbacks of JSON include lack of type definition and some sort of DTD.
 
   "that was short": "And, you're done. You now know everything JSON has to offer."
 }
+
+// Object JSON FROM String
+// The JSON.parse() method parses a string as JSON, optionally transforming the value produced by parsing.
+
+// Sintax:
+// JSON.parse(text[, reviver])
```
vendethiel: what is github doing here..?

raphaelbn:
it's so strange.. I don't know why github has colored this word.. @vendethiel 

do you have any idea?

json.html.markdown
awalGarg: @vendethiel This is github's markdown parser interpreting `[, reviver]` as bad JSON. JSON doesn't have comments, and the `[, reviver]` is interpreted as an array literal, but since `,` is not a valid first token inside an array literal, it borks the parsing. This entire comment block would be best be separated from the json block.

awalGarg:
Not sure if this is something to include in the JSON doc. JSON.parse/stringify is strictly an ECMAScript feature, and JSON is _not_ tied to ECMAScript. Most other languages include out of the box support for JSON parsing (Python, PHP, Ruby, Go, etc.).

This, if not already documented, belongs to the JS doc instead, IMHO.

PR #1632, opened 2015-10-18
More detail

erich1510:


java.html.markdown
```diff
@@ -686,7 +686,6 @@ public abstract class Mammal()
         return true;
     }
 }
-```
```
onebree: You should not remove this, as it closes the markdown formatting for code (started on line 18)

c++.html.markdown
```diff
@@ -215,7 +215,7 @@ cout << myString + myOtherString; // "Hello World"
 
 cout << myString + " You"; // "Hello You"
 
-// C++ strings are mutable and have value semantics.
+// C++ strings are mutable and have value semantics. This differs from Java.
```
geoffliu: This differs from pretty much any other modern language.

PR #1635, opened 2015-10-18
Added more resources to LaTex

andela-borazulume:


voltnor:
The project homepage is a good resource. CTAN should probably also be added. I'm torn on the tutorials. There are an infinite number of them on the internet, and of variable quality. I don't know of a curated LaTex tutorial list, but the tex stack exchange has a list: http://tex.stackexchange.com/questions/11/what-are-good-learning-resources-for-a-latex-beginner

Maybe we could put the resources in a form like the Java learnXinY with sections like "More on LaTex" (project homepage, wikibook, CTAN), "Tutorials", and "Books"?

PR #1636, opened 2015-10-18
[bash/en]...Added info on changing directories

viv1:


vendethiel:
Thanks & sorry for the delay.

PR #1638, opened 2015-10-18
[php/en] Added variable amount of parameters as of php 5.6+

xssc:
added use of '...' for variable amount of arguments


adambard:
Thanks!

PR #1639, opened 2015-10-19
[php/en] Add more examples and explanations

labrack:
Added some clarifying examples to sections on echo(), constants, and
cleaned up formatting on others. Added further explanation about the
spaceship operator.


adambard:
Thanks!

PR #1652, opened 2015-10-19
Updated official python docs link.

MineDoubleSpace:
Updated official python docs link to always use the latest version of python 2.x


adambard:
Good plan, thanks.

PR #1654, opened 2015-10-19
[matlab-en] Figure handle and some text issues fixed

claudsonm:
While I was translating the file (#1628) I noticed these little mistakes, so there we go.

Filename also added to the markdown.


adambard:
Thanks!

PR #1685, opened 2015-10-20
[xml/en] Grammar, formatting. Made more 'inlined'.

duci9y:
Also fixed incorrect code and added information about XML schemas. Also added links and references.


Spurlow:
Thanks!

PR #1693, opened 2015-10-20
[paren/pt-br] Translation to Brazillian Portuguese

claudsonm:
Content translated and revised.

PR #1695, opened 2015-10-20
[paren/en] Fixing quotes inside comments and typing error

claudsonm:
While I was translating the paren file (#1693) I noticed these things.

I think that these changes will improve the content. And you can see [here](https://en.wikipedia.org/wiki/Syntactic_sugar) the correct name definition.


Oire:
Thanks!

PR #1702, opened 2015-10-20
Added How to swap 2 numbers without using 3rd variable

samsruti:


vendethiel:
meh, only works with integer and is not generally good practice.

PR #1703, opened 2015-10-20
Modify the function composition example to avoid two 5's

Gregly:
The example: (5 + 10) * 5, might cause a slight confusion regarding which 5 is the function argument. This update should make it very straightforward.


vendethiel:
That's fair. Thanks!

PR #1706, opened 2015-10-21
Explicitly show that "new" is a keyword that instantiates an class obj.

bharathkkb:


vendethiel:
I think that's plenty good enough :)

PR #1713, opened 2015-10-21
[java/en] Static code block

perlilja:


adambard:
I had no idea that existed. Thanks!

PR #1724, opened 2015-10-21
[nl/Markdown] Add dutch translation of markdown

HetIsNiels:
Add dutch translation of markdown


vendethiel:
thanks&sorry for the delay!

PR #1733, opened 2015-10-22
[tmux] Updating for version 2.1

matthiaskern:


vendethiel:
Sorry for the delay, but I'm afraid we reserve the "contributors" tag for significant contribution0s

PR #1736, opened 2015-10-22
[php/en] Un-beef a comment block closing

labrack:
Somebody beefed a closing on the 'Late Static Binding' Comment block, causing the next 18 or so lines to be included in the comment instead of parsed for syntax properly.


vendethiel:
whoops! Nice catch, thanks for the PR!

PR #1742, opened 2015-10-23
[python/en] [python3/en] Added string length

thesourabh:
Knowing how to find the length of a string would be pretty beneficial for new users.

Didn't use a variable since variables are only introduced in the next section, so I continued from the previous example string.


thesourabh:
Updated for python3 as well.

thesourabh:
@vendethiel Updated for python 3 as well

vendethiel:
"merged" manually. Sorry for the delay (did that from the web interface – this had conflict)

PR #1746, opened 2015-10-23
Pythonstatcomp spanish translation

damasosanoja:
This is the spanish translation of Python statical components


damasosanoja:
Sorry just saw my error, correcting now.

PR #1757, opened 2015-10-23
[make/en] Example of typical usage of phony target

juusaw:
An example clarifying the use of phony target 'clean'


make.html.markdown
```diff
@@ -71,7 +71,10 @@ maker:
 # A rule with a dependency on a phony target will always run
 ex0.txt ex1.txt: maker
 
+# Phony targets are used to perform actions to source.
```
vendethiel: that's a bit vague, I think?

PR #1758, opened 2015-10-23
Some minor fixes

patrikja:
I just noted that the example claiming that (add 10) is the same as (+10) was wrong. (A detail - it should be (10+) to match the argument order.) Then I just continued down making a few similar fixes and terminology updates. 
Kind regards,
  Patrik

PS. I've been teaching [Advanced Functional Programming](http://www.cse.chalmers.se/edu/course/afp/) (in Haskell) for a few years at Chalmers.


ignlg:
Please, merge this PR.

vendethiel:
Thanks @patrikja !

PR #1772, opened 2015-10-24
corrected typo error

himanshu81494:
error in typo

"muster" in place of "master"


geoffliu:
I believe this has been fixed in another PR. Thanks for sending though!

PR #1775, opened 2015-10-24
[python/en] Added <> to boolean operators

Dor-Ron:
It's a lesser known and less common part of the language, but much like SQL <> can be used to see if 2 objects are dissimilar in Python 2.

### DIFF
--- a/python.html.markdown
+++ b/python.html.markdown
@@ -101,6 +101,8 @@ not False  # => True
 1 <> 1  # => False
 2 <> 1  # => True
 
+# Note the <> is deprecated and will not work for python 3.x
+
 # More comparisons
 1 < 10  # => True
 1 > 10  # => False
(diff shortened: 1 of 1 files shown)


payet-s:
Be careful, this syntax is deprecated in Python2 and incorrect in Python3.
You should at least mention the deprecated aspect of this syntax alternative.

Otherwise, it's always good to know, specially when you're reading someone else code that might use it.

Dor-Ron:
You're right, comment added!

vendethiel:
I'm not sure we should show deprecated features.

payet-s:
It might be good to know. Especially because this syntax is used in others languages like Pascal and it can be present in someone else code. At least, you know that it exists even if you don't use it often.

geoffliu:
I have never seen this syntax used, and with the deprecation, it's going to be less relevant.

PR #1776, opened 2015-10-24
Update javascript-es.html.markdown

imagentleman:
Changed:

1. The conjunction, to match the original meaning (in the English article) and the context. The original word was "though", which better resembles "however" ("sin embargo" in spanish). "Aunque" is commonly used to counter something (expressed in the same sentence), closer to the way someone would use "although". The way the translation is written, it's more of a plain statement (with the conjunction at the beginning of the clause), so "Sin embargo" (however) or "Pero" (but) are better alternatives than "Aunque".
2. An Uppercase [sic] : ) typo.
3. Use of strict triple equals to match the original English article.

### DIFF
--- a/es-es/javascript-es.html.markdown
+++ b/es-es/javascript-es.html.markdown
@@ -16,7 +16,7 @@ con Java para aplicaciones más complejas. Debido a su integracion estrecha con 
 web y soporte por defecto de los navegadores modernos se ha vuelto mucho más común 
 para front-end que Java.
 
-Aunque JavaScript no sólo se limita a los navegadores web: Node.js, Un proyecto que proporciona un entorno de ejecución independiente para el motor V8 de Google Chrome, se está volviendo más y más popular.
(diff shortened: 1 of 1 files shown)


es-es/javascript-es.html.markdown
```diff
@@ -81,12 +81,12 @@ false;
 !false; // = true
 
 // Para comprobar una igualdad se usa ==
```
vendethiel: please update the operator here as well

es-es/javascript-es.html.markdown
```diff
@@ -81,12 +81,12 @@ false;
...
-1 == 1; // = true
-2 == 1; // = false
+1 === 1; // = true
+2 === 1; // = false
 
 // Para comprobar una desigualdad se usa !=
```
vendethiel: here too :)

vendethiel:
Sorry for the delay.

imagentleman:
Created an updated pr to address the comments: #2267.

PR #1780, opened 2015-10-25
[java/en] String concatenation added.

JaXt0r:


java.html.markdown
```diff
@@ -99,7 +99,21 @@ public class LearnJava {
         System.out.println(fooString);
         System.out.println(barString);
         System.out.println(bazString);
-
+        
+        // String Building
+        // #1
```
geoffliu: Numbering is a bit unnecessary.

java.html.markdown
```diff
@@ -99,7 +99,21 @@ public class LearnJava {
...
+        String plusConcatenated = "Strings can " + "be concatenated " + "via + operator";
+
+        // #2
+        StringBuilder builderConcatenated = new StringBuilder();
```
geoffliu: `StringBuilder` is rather rarely used. I think I run into a good use case once every three years. We could include it if you feel strongly, but at least also include an explanation of its advantage over using just `+`. My vote is to take this out though.

geoffliu:
@JaXt0r Thanks! A few comments, and we can merge this after you address them.

java.html.markdown
```diff
@@ -99,7 +99,21 @@ public class LearnJava {
...
+        builderConcatenated.append("You ");
+        builderConcatenated.append("can use ");
+        builderConcatenated.append("the StringBuilder class.");
+        System.out.println(builderConcatenated.toString());
+        
+        // #3
+        String.format("%s may prefer %s", "Or you", "String.format()");
```
geoffliu: Include the result in comments?

java.html.markdown
vendethiel: Nah, it might be good to teach people not to abuse `+`, because of performance implications.

However, the mention of `StringBuffer` (the thread-safe version) might not be necessary

JaXt0r:
Thanks for the reply. I added some comments and pushed them.

vendethiel:
There's a few typo I need to fix when I merge this, so I'll wait until I have a computer at hand.

vendethiel:
Thanks. There were still conflicts, so I just merged it manually it, instead of sending it back one more time.

PR #1789, opened 2015-10-25
Update java.html.markdown, use Here instead of This in comments

iandioch:
Clear up confusion by saying "here is" instead of "this is", to clear up confusion (as the "this" keyword is used for the first time here, it's easy to misinterpret the comment)


vendethiel:
I don't think it's an actual issue, especially since it's uppercased :) (also, sorry we took so long to answer, that made it conflict). Thanks still!

PR #1804, opened 2015-10-26
Change curly braces statement

scoops:
The '$$' before the braces was a bit confusing. I added a couple lines that show that you have options on where the '$' goes when it comes to curly braces in strings.


vendethiel:
Thanks & sorry for the delay!

PR #1808, opened 2015-10-27
[fsharp/en] correct a few simple mistakes

sholland1:
change github flavored markdown programming language to fsharp
correct typos

### DIFF
--- a/fsharp.html.markdown
+++ b/fsharp.html.markdown
@@ -16,7 +16,7 @@ The syntax of F# is different from C-style languages:
 
 If you want to try out the code below, you can go to [tryfsharp.org](http://www.tryfsharp.org/Create) and paste it into an interactive REPL.
 
-```fsharp
+```csharp
 
 // single line comments use a double slash
 (* multi line comments use (* . . . *) pair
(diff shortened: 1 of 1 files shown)


vendethiel:
I'm not sure our parser supports fsharp...

sholland1:
F# was heavily influenced by OCaml, and it looks the [OCaml](http://learnxinyminutes.com/docs/ocaml/) page is being parsed correctly. I can change it to use that if the parser doesn't support F#.

sholland1:
Ok, so I did some digging, and I am guessing the parser you guys are using is [Rouge](https://github.com/jneen/rouge), since it used by the Middleman extension [middleman-syntax](https://github.com/middleman/middleman-syntax).

It appears that there is currently no support for F# in Rouge.

I also found an [online syntax highlighter](http://rouge.jneen.net/) that uses Rouge and pasted the F# code that we have into it. I tried using the OCaml and SML highlighting modes, but because neither of these languages have `//single line comments`, the highlighting for those comments was wrong.

I will change the language back to C#.

vendethiel:
Right. Thanks :)

PR #1825, opened 2015-10-27
[Brainfuck/es] update translation

julianaortga:
Hello.
Update translation and minor syntax fix.


vendethiel:
It seems this is "outdated" because of the delay. Sorry about that...

PR #1735, opened 2015-10-22
Add PCRE #1561

ssd532:
For learning regular expressions a brief guide to PCRE syntax is added.

PR #1826, opened 2015-10-27
Add an example of trap command 

ssd532:
`trap` is a very important command to intercept a fatal signal, perform cleanup, and then exit gracefully. It needs an entry in this document.

Here a simple and most common example of using `trap` command i.e. cleanup upon receiving signal is added.


ssd532:
I am sorry. Changes from #1735 are propagated here. I have done something wrong during forking,  merging and while creating PR. What can we do about it?

vendethiel:
Thanks. I'll just merge both as the same time :-).

Sorry for the delay.

PR #1834, opened 2015-10-28
checking css and spacing

farque65:


bash.html.markdown
```diff
@@ -31,75 +32,93 @@ echo Hello world!
 echo 'This is the first line'; echo 'This is the second line'
 
 # Declaring a variable looks like this:
-VARIABLE="Some string"
```
onebree: I thought the convention was to use UPPER_CASE for bash variables.

onebree:
@farque65 I think this pull request is too large for anyone to review timely. I mean, it may be best to just create a new branch from master (currently), and reapply your commits (not the ones from others, seen above).

vendethiel:
I think @onebree summed it up then...

PR #1835, opened 2015-10-28
[ruby-ecosystem/en] Adding Some Popular Ruby Gems

brgaulin:
Seeing some of the popular gems are a good example of the Ruby Ecosystem. I didn't want to add too many, so I based my choices on ones I have seen frequently as well as what rubygems had for stats.


onebree:
I don't think "popular gems" are necessary to include. There are already awesome-list sites dedicated to just that. Also, a JSON library is already included (in 2.2.1 _at least_).

brgaulin:
:+1:

PR #1836, opened 2015-10-28
[javascript/en] "wan't" -> "want"

corbanmailloux:


adambrenecki:
Huh, how did that get in there?

Thanks!

PR #1841, opened 2015-10-28
Removed whitespacing all over

imjching:
[java/en] added an alternative name for For-Each loop: Enhanced for Loop;
[typescript/en] fixed some markdown syntax;
removed whitespacing for the remaining files


geoffliu:
@imjching Thanks for the edit. When you have 22 files in a PR though, it's likely that merge conflicts will arise. If you feel strongly about removing white spaces (and I agree with you that we should), please open PRs for one file at a time. There's a lower chance that there will be conflicts.

Issue #1253, opened 2015-09-28
[javascript] ES6


m90:
There **needs** to be ES6. Anyone working on that already?

I'm not too sure if this should be treated as a separate language or go as an addendum to JavaScript (will probably break the runnable example rule)?

adambard:
Yeah, it should definitely be separate (as with python/python3)

tech4david:
Just started working on this

awalGarg:
@adambard I think the python/python3 case won't be good here since Python 3 breaks backwards compatibility quite enormously, while ES6 is 100% backwards compatible with ES5 (leave for some minor edge cases rarely ever encountered IRL which nobody cares about).

Although I 100% agree that ES6 is one major update (http://i.imgur.com/X53WgY6.png), it'd still not be that much of a big addition to the current JS doc at learnxiny because of the conciseness that learnxiny docs aim for. For instance, most of the new additions to Array.prototype and String.prototype wouldn't make for much place in the doc just like `Array.prototype.forEach` and ilk aren't documented in the current ES5 doc.

Addition of the new syntax (arrow functions, destructuring etc) and exotic objects (like generators and proxies) should do, IMHO.

Thoughts?

m90:
Agree with @awalGarg here. DRY. A concise addition to the existing ES5 one should be better than duplicating thing like simple math.

levibostian:
@awalGarg explains it really well. I agree keeping it all in 1 doc.

tech4david:
Sorry for taking so long, finally got round to doing it

adambrenecki:
There's two good PRs adding ES2015 already, and my intention is to do a bit of editing and merge the best of both; I'd forgotten about this but now that I've been reminded I'll do it this weekend.

Originally when those PRs came up I agreed with @adambard that ES2015 deserved its own article, but now that all modern browsers support nearly all of it I think it's appropriate to merge it into the original, indicating which parts are ES2015-only as appropriate.

adambard:
Yes, I agree with this -- we can start adding features to the JS article. I guess I'll try to dredge up one of the old PRs and see if they still merge

adambrenecki:
I'm going to edit the best bits of both of them into the current document manually this weekend; given how long it's been that's probably a better option than unleashing git merge onto it.

[...]

PR #1844, opened 2015-10-28
Add ES6 stuff to JavaScript file

techdavid:
See #1253

### DIFF
--- a/javascript.html.markdown
+++ b/javascript.html.markdown
@@ -3,6 +3,7 @@ language: javascript
 contributors:
     - ["Adam Brenecki", "http://adam.brenecki.id.au"]
     - ["Ariel Krakowski", "http://www.learneroo.com"]
+    - ["David Bailey", "https://tech4david.github.io/"]
 filename: javascript.js
 ---
 
@@ -134,7 +135,7 @@ undefined; // used to indicate a value is not currently present (although
 // Note that 0 is falsy and "0" is truthy, even though 0 == "0".
 
 ///////////////////////////////////
-// 2. Variables, Arrays and Objects
+// 2. Variables, Strings, Arrays and Objects
 
 // Variables are declared with the `var` keyword. JavaScript is dynamically
 // typed, so you don't need to specify type. Assignment uses a single `=`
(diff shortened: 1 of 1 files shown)


adambrenecki:
This looks good, but IMO it might be helpful to put a sentence about ES6 in the preamble at the top, so people know what "ES6" means; it could also link to things like Babel.

Also, just an idea: what if each ES6 feature had something easily scannable at the beginning of its leading comment, e.g.:

```js
// [ES6 only] You can specify a default argument to use if one isn't passed in.
```

tech4david:
> what if each ES6 feature had something easily scannable

You can <kbd>ctrl</kbd><kbd>f</kbd> "ES6 environments" but I suppose I could write it like that instead to make it clearer at a glance. And yeah, I guess I should add something to the introduction about it

adambard:
I don't think the world is ready to just have ES6 dropped inline into the general "Javascript" file yet. I like @adambrenecki's suggestion, or else just lumping the ES6 stuff together in one, easily-ignored pile for the benefit of those who can't use ES6 yet.

vendethiel:
we also have another ES6-oriented PR.

ryanplant-au:
If (when?) it's included, should it be called ES6 or ES2015? ES6 has been more popular, but ES2015 is the preferred official use, and it's what I'm seeing on book covers lately.

The ES2016/ES7 standard was published this month, on the first anniversary of ES2015/ES6. Is one year a long enough wait before adding it?

vendethiel:
I think it should be more about browser share?

ryanplant-au:
Well, browser share's already pretty strong. The big ES6/2015 features like rest parameters, let, const, proxy, and arrow functions are all supported by three of the big four current major browsers (Firefox, Edge, and Chrome -- only Safari lags) according to CanIUse.

vendethiel:
Yeah, the time might be now.

vendethiel:
Still unsure what to do with this one and #1905.

vendethiel:
Uhhh. That closed itself automatically and now I can't reopen...

adambard:
The source repo was deleted. Guess we're out of luck.

techdavid:
@adambard my bad, just a sec and I'll do it again

techdavid:
@adambard sorry bout that #2491

PR #1847, opened 2015-10-28
Add one more string formatting example

cnu:
String templates doesn't need to have a number of keyword to format.


Spurlow:
Thanks!

PR #1849, opened 2015-10-28
correcoes ortograficas e adicao de alguns exemplos

caionitro:
Hello, 
I added an example of echo and print and added one ; was missing on line 69.

(sorry my bad english)


vendethiel:
Thanks & sorry for the delay!

PR #1860, opened 2015-10-28
[C/en] Added tidbit about converting cmdline args to numerical values

luehm:
[C/en] Thought it would be appropriate to show people how to convert
a string command-line argument into it's numerical form.

Didn't know if it was appropriate to show basic error checking
or not...

### DIFF
--- a/c.html.markdown
+++ b/c.html.markdown
@@ -63,12 +63,22 @@ int main(void) {
   // your program
 }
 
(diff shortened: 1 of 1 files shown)


vendethiel:
Going to close this, as `sscanf` is mostly considered dangerous (buffer overruns and all).

PR #1864, opened 2015-10-28
[csharp/en] Add string interpolation

bigbash:
Added example of using string interpolation


vendethiel:
Thanks&sorry for the delay.

PR #1866, opened 2015-10-28
[javascript/en] Adding a small blurb to extend upon string concatenation

Pink401k:
Small change, but I think showing that `int + strings = string` is good, since that's somewhat atypical.

Didn't add myself to contributors since it's so minor.


adambrenecki:
:+1:

PR #1869, opened 2015-10-28
[C/en] Added emphasis on importance of NULL terminated strings.

luehm:
After teaching students strings for a while, I've found that this
concept always seems to cause much confustion

### DIFF
--- a/c.html.markdown
+++ b/c.html.markdown
@@ -76,7 +76,7 @@ int main (int argc, char** argv)
   ///////////////////////////////////////
   // Types
   ///////////////////////////////////////
-  
+
   // All variables MUST be declared at the top of the current block scope
   // we declare them dynamically along the code for the sake of the tutorial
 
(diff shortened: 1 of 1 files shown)


c.html.markdown
```diff
@@ -162,6 +162,8 @@ int main (int argc, char** argv)
 
   // Strings are just arrays of chars terminated by a NULL (0x00) byte,
   // represented in strings as the special character '\0'.
+  // NULL characters are vital to functions that operate on strings,
```
geoffliu: Reads a bit odd when together with the next sentence. Maybe combine them like "We don't have to ..., but if you're constructing one in code, be sure to include them"? Also, while we're on the subject, definitely mention that you should have an extra byte when allocating space for strings. That seems to trip people up as well.

geoffliu:
@luehm Also seems like there are merge conflicts. Could you resolve?

PR #1872, opened 2015-10-28
[C/en] Added tidbit about fall-though in switch statements.

luehm:
Another pitfall, as not all languages have fall-through in switches.


geoffliu:
Yes, thank you!

PR #1875, opened 2015-10-29
[xml/nl] Add Dutch XML translation

frvge:
Dutch version of XML

### DIFF
--- a/nl-nl/xml-nl.html.markdown
+++ b/nl-nl/xml-nl.html.markdown
@@ -1,0 +1,134 @@
+---
+language: xml
+filename: learnxml-nl.xml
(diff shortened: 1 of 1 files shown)


frvge:
What's taking so long? Simple translations should be just a simple merge. =/

vendethiel:
Thanks & sorry for the delay! what usually happens is that we leave the PR open for a bit, in case a native speaker of that language can help us review it, and we merge it once we're sure it's been reviewed carefully. We don't however have a dutch speaker around in the team, so the PR slipped. Just a nudge and we'll usually merge it!

Thanks & sorry for the delay again :).

PR #1876, opened 2015-10-29
Output example had a value more

paulohrpinheiro:
The third line of the sample is invalid, only two values are generated.


vendethiel:
Thanks&sorry for the delay.

PR #1887, opened 2015-10-29
[c++/en] Added Lambda Expressions and Range for

bk2dcradle:
I added how to use "Lambda Expressions" and "Range For" examples.


c++.html.markdown
```diff
@@ -801,6 +801,106 @@ void doSomethingWithAFile(const std::string& filename)
...
 
+///////////////////////////////////////
+// Lambda Expressions (C++11 and above)
+///////////////////////////////////////
+
+// lambdas are a convenient way of defining an anonymous function
+// object right at the location where it is invoked or passed as 
+// an argument to a function.
+
+// Example consider sorting a vector of pairs using the second
```
geoffliu: "For example, consider..."

c++.html.markdown
```diff
@@ -801,6 +801,106 @@ void doSomethingWithAFile(const std::string& filename)
...
+vector<pair<int, int> > tester;
+tester.push_back(make_pair(3, 6));
+tester.push_back(make_pair(1, 9));
+tester.push_back(make_pair(5, 0));
+
+// Pass a lambda expression as third argument to the sort function
+// sort is from the <algorithm> header
+
+sort(tester.begin(), tester.end(), [](const pair<int, int> &lhs, const pair<int, int> &rhs) {
+		return lhs.second < rhs.second;
```
geoffliu: I believe the rest of this document uses the `typename& var_name` convention. Also, spaces instead of tabs. Now that I look at it, the spaces are quite inconsistent... heh. But still, let's keep to spaces.

c++.html.markdown
```diff
@@ -801,6 +801,106 @@ void doSomethingWithAFile(const std::string& filename)
...
+
+///////////////////////////////
+// Range For (C++11 and above)
+///////////////////////////////
+
+// You can use a range for loop to iterate over a container
+int arr[] = {1, 10, 3};
+
+for(int elem: arr){
+	cout<<elem<<'\n';
```
geoffliu: Spaces around the `<<` operator, and also, use `endl` rather than `\n`

c++.html.markdown
```diff
@@ -801,6 +801,106 @@ void doSomethingWithAFile(const std::string& filename)
...
+// [] in the lambda is used to "capture" variables.
+// For Example:
+
+vector<int> dog_ids;
+// number_of_dogs = 3;
+for(int i = 0; i < 3; i++){
+	dog_ids.push_back(i);  
+}
+
+int weight[3];
```
geoffliu: Would be simpler to just do `{30, 50, 10}`, no?

c++.html.markdown
```diff
@@ -801,6 +801,106 @@ void doSomethingWithAFile(const std::string& filename)
...
+weight[0] = 30, weight[1] = 50, weight[2] = 10;
+
+// Say you want to sort dog_ids according to the dogs' weights
+// So dog_ids should in the end become: [2, 0, 1]
+
+// Here's where lambda expressions come in handy
+
+sort(dog_ids.begin(), dog_ids.end(), [weight](const int &lhs, const int &rhs) {
```
geoffliu: Better to capture `weight` here by reference. In fact, could you expand on the capture list? Just quick one-line explanations on the syntax for capture by value, by reference, `[=]`, and `[&]`

c++.html.markdown
```diff
@@ -801,6 +801,106 @@ void doSomethingWithAFile(const std::string& filename)
...
+		return weight[lhs] < weight[rhs];
+	});
+// Note we captured "weight" in the above example.
+
+// lambda are really useful for the case of structs
```
geoffliu: From here until the end of the section, I'm not sure if it's necessary. This is meant to be a reference page, not a full tutorial, so long sections of use cases might not be appropriate. Also, comparator function objects are quite common in other languages, so it wouldn't be a new concept to most people.

c++.html.markdown
```diff
@@ -801,6 +801,106 @@ void doSomethingWithAFile(const std::string& filename)
...
+
+// You can use a range for loop to iterate over a container
+int arr[] = {1, 10, 3};
+
+for(int elem: arr){
+	cout<<elem<<'\n';  
+}
+
+// You can use "auto" and not worry about the type of the elements of the container
+// Caveat: Don't assign inside the range for loop
```
geoffliu: I'm not sure about this blanket warning of "don't assign in side the loop." The reason it's not working is that you declared the variable as `auto elem`, not because it's inside a loop; there's nothing special about being in the loop. We should make it clear that when you do `int elem` or `auto elem`, the value is *copied* into `elem`. Doesn't make a huge difference here since it's just an int, but it would matter for objects. I actually prefer to have `const auto&` in the range loops unless there's a good reason not to.

geoffliu:
@bk2dcradle Thanks for the write-up. I've been meaning to get around to writing a section on lambda, but didn't really have the time. This is big help :)

Just a few comments, most of them minor. The loop declaration thing is a bit more involved, I do hope you agree with me and revise. Also, seems like there are merge conflicts with master, could you resolve?

bk2dcradle:
Resolved the Merge Conflicts. And I agree with your remark regarding the Range For loop. I was really short on time when I was writing this a few months back. 
I have made the necessary changes, let me know if anything is left.

c++.html.markdown
```diff
@@ -820,6 +820,73 @@ std::map<Foo, int, compareFunction> fooMap;
...
+vector<pair<int, int> > tester;
+tester.push_back(make_pair(3, 6));
+tester.push_back(make_pair(1, 9));
+tester.push_back(make_pair(5, 0));
+
+// Pass a lambda expression as third argument to the sort function
+// sort is from the <algorithm> header
+
+sort(tester.begin(), tester.end(), [](const pair<int, int>& lhs, const pair<int, int>& rhs) {
+		return lhs.second < rhs.second;
```
geoffliu: Tabs still here.

c++.html.markdown
```diff
@@ -820,6 +820,73 @@ std::map<Foo, int, compareFunction> fooMap;
...
+
+// Say you want to sort dog_ids according to the dogs' weights
+// So dog_ids should in the end become: [2, 0, 1]
+
+// Here's where lambda expressions come in handy
+
+sort(dog_ids.begin(), dog_ids.end(), [&weight](const int &lhs, const int &rhs) {
+		return weight[lhs] < weight[rhs];
+	});
+// Note we captured "weight" by reference in the above example.
```
geoffliu: Could you add a few words about what the other types of captures? Or at least leave a comment like "TODO: add other types of captures"

c++.html.markdown
```diff
@@ -820,6 +820,73 @@ std::map<Foo, int, compareFunction> fooMap;
...
+
+// lambda are really useful for the case of structs
```
geoffliu: This paragraph doesn't make much sense if you take out the rest of the section. Let's get rid of it too. I think the examples given above are sufficient.

geoffliu:
@bk2dcradle A few comments left over, thanks for putting up with my pedanticalness.

bk2dcradle:
I have tried to explain other types of captures as briefly as possible. Again, let me know if anything is left. I don't mind the pedanticalness, it will only improve the Doc.

geoffliu:
Yup, thanks!

PR #1889, opened 2015-10-29
[sass/cs] Czech translation for Sass

MichalMartinek:
Czech translation for Sass

### DIFF
--- a/cs-cz/sass.html.markdown
+++ b/cs-cz/sass.html.markdown
@@ -1,0 +1,439 @@
+---
+language: sass
+filename: learnsass.scss
+contributors:
+  - ["Laura Kyle", "https://github.com/LauraNK"]
+  - ["Sean Corrales", "https://github.com/droidenator"]
+translators:
+  - ["Michal Martinek", "https://github.com/MichalMartinek"]
+
+---
+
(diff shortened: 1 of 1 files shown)


cs-cz/sass.html.markdown
```diff
@@ -0,0 +1,439 @@
+---
+language: sass
+filename: learnsass.scss
+contributors:
+  - ["Laura Kyle", "https://github.com/LauraNK"]
+  - ["Sean Corrales", "https://github.com/droidenator"]
+translators:
+  - ["Michal Martinek", "https://github.com/MichalMartinek"]
+
```
adambard: Could you add a `lang` line? I know it's a bit weird, but the term "language" gets a bit overloaded these days.

```
lang: cs-cz
```

cs-cz/sass.html.markdown
```diff
@@ -0,0 +1,439 @@
+---
+language: sass
+filename: learnsass.scss
```
adambard: Could you also change the filename to "learnsass-cz.scss" or some such thing? Otherwise there can be collisions.

MichalMartinek:
Sorry for my late reply. It should be okay now.

adambard:
looks good, thanks!

PR #1893, opened 2015-10-29
add a D resource

yihongang:
added "Programming in D" as a resource


vendethiel:
link doesn't load here

yihongang:
Hmm even if 0 dollars is a valid price?

vendethiel:
That didn't work for me. Also, that's not fair

PR #1896, opened 2015-10-29
tidy inline comments

yihongang:
align and visually separate inline comments from code


geoffliu:
Eh I'm going to close this one. This kind of alignment is a nightmare to maintain, and it doesn't add much to readability.

PR #1899, opened 2015-10-29
[Angular/en-en]

WalterC87:
Create an AngularJS Tutorial


adambard:
I think you'll want to point this at the master branch :)

PR #1910, opened 2015-10-30
[php/en] echo fix and cleanup

definitelyliam:
Line 159 - echo isn't a function, print() is
Line 337 - unneeded semicolon (it's a pre-test loop)


adambard:
Thanks!

PR #1914, opened 2015-10-30
Add return for ruby

alcastaneda:
Just added information about return in ruby

### DIFF
--- a/ruby.html.markdown
+++ b/ruby.html.markdown
@@ -127,7 +127,7 @@ print "I'm printing!"
 
 #You can use return in ruby
 #but it is mostly unecessary since
-#ruby inherently returns the output
+#ruby inherently returns the output of functions
 
 
 # Variables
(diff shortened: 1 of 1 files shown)


ruby.html.markdown
```diff
@@ -113,7 +114,7 @@ placeholder = 'use string interpolation'
 'hello ' * 3 #=> "hello hello hello "
 
 # Append to string
-'hello' << ' world' #=> "hello world"
+'hello' << '  world' #=> "hello world"
```
geoffliu: Why the extra space

ruby.html.markdown
```diff
@@ -124,6 +125,11 @@ puts "I'm printing!"
 print "I'm printing!"
 #=> I'm printing! => nill
 
+#You can use return in ruby
```
geoffliu: This is an odd place for this explanation. I believe there's already a line about this in the section about functions.

PR #1918, opened 2015-10-30
[json/pt-br] Improved the JSON document translation

ToFran:
There were some typos, and added more information (translated from the en version).


adambard:
Obrigado!

PR #1922, opened 2015-10-31
Added a couple functions to the matlab doc.

gsenesac:
Added some entries about:
 1) Generating an array with increment other than 1
 2) More concise transpose operators
 3) A couple more plotting options
 4) Some complex number specific math functions
 5) the find() function, which I use quite a bit and think others might find useful


adambard:
Thanks!

PR #1923, opened 2015-10-31
[csharp/en] added and normalized curly braces

jason-collier:
It's generally considered a best practice in C# to include curly braces, even when they're not necessary, so I've added them wherever they were left out. 

Also, where the first curly brace appeared in the first line of a declaration, I moved the brace down to its own line, which is more typical of C#'s style.

### DIFF
--- a/csharp.html.markdown
+++ b/csharp.html.markdown
@@ -8,6 +8,7 @@ contributors:
     - ["Wouter Van Schandevijl", "http://github.com/laoujin"]
     - ["Jo Pearce", "http://github.com/jdpearce"]
     - ["Chris Zimmerman", "https://github.com/chriszimmerman"]
+    - ["Jason Collier", "https://github.com/jason-collier"]
 filename: LearnCSharp.cs
 ---
 
@@ -255,12 +256,16 @@ on a new line! ""Wow!"", the masses cried";
             {
                 // Start iteration 100 times, fooDoWhile 0->99
                 if (false)
+                {
                     continue; // skip the current iteration
+                }
 
                 fooDoWhile++;
 
                 if (fooDoWhile == 50)
+                {
                     break; // breaks from the loop completely
+                }
 
             } while (fooDoWhile < 100);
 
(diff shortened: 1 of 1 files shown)


adambard:
This feels like a really subjective change, do you have some style guide or something you can cite?

jason-collier:
Hi!

I agree that it is a subjective issue but only insofar as we're dealing with an individual and their own code. Since this document is meant for general consumption, and those learning from it may be working with others, I figure it's best to take the approach with the least potential for confusion and error, which means adding the braces.

As far as style guides or references go, any Google search for "C# curly braces" will bring up a ton of threads discussing the issue. There are plenty of opinions both ways so anything I show you in that regard would inevitbly be prejudiced towards my own view on the matter. (Such is the nature of "evidence" on the internet.) I'd rather just give my reasoning and hope you find it reasonable as well.

* People coming from other languages will not be confused by the presence of curly braces but they may be confused by the *absence* of them in what may appear to be random locations. Therefore, in this situation, it's better to include them.

* There are only a few situations where the curly braces can be left out. In all other situations they're required. By always using them, the code remains consistent. One thing both sides of this matter agree upon is *remaining consistent* with whatever choice has been made. I argue we reamin consistent *with the rest of C#* by using braces everywhere.

* Leaving the braces out creates the potential for problems that do not exist when they're included, including problems for people maintaining the code. ([example](http://stylecop.soyuz5.com/SA1503.html))  If an individual doesn't mind this risk, that's fine, but in a group environment, or when considering it objectively, I think it makes sense to take the safer route.

geoffliu:
@jason-collier I would agree with you on one class of change here, and that is to align the opening brace on a new line where it isn't already. The C# community does seem to lean that way.

The addition of braces of single-line blocks, as @adambard mentioned, is subjective. I think for tutorial purposes, it is better to show that these braces can be omitted. This behavior is consistent with many other related languages, so I don't think it'll be confusing. 

Lastly, although we don't have a strict guideline on what counts as authorship for these articles, usually we only add people's name at the top of the page if they have contributed significant sections. Again, we don't know exactly where the dividing line is, but I believe cleanup patches are not quite there.

vendethiel:
Yeah, I'm just gonna go ahead and close this, as long as the file is consistent :).

Thanks still!

PR #1925, opened 2015-10-31
[Objective-C/en] Adds description of how to define generic classes in Xcode 7.0

benpious:
As the title. Covers the basic case.


adambard:
Cool, ty!

PR #1932, opened 2015-10-31
[haskell/en]: Fix !! operator

xou:
- Use a finite list, as infinite lists are not introduced yet
- Use a list starting from 1 instead of 0, to make it obvious that
  the element returned comes from the list (and is not the second
  argument to !!).


adambard:
Thanks!

PR #1935, opened 2015-10-31
[haskell/de] Synchronized with english version

xou:
Added changes from a7b8858 to HEAD (b5cb14703) to the
german version, where appropiate.


adambard:
Danke!

PR #1939, opened 2015-10-31
[julia/en] Add information about keyword arguments in functions section of julia

pranitbauva1997:
There are many programmers who are unaware of keyword arguments and thus a short info about keyword argument is useful.


adambard:
I think the code example explains this adequately for our purposes. Thanks for your contribution though!

PR #1941, opened 2015-10-31
[haskell/de] [yaml/de] Fix umlauts

xou:
Replaced ae/oe/ue with ä/ö/ü where appropriate.

I did a "grep" in the de-de directory, i've only found
problems in haskell and yaml files.


adambard:
Thanks!

PR #1943, opened 2015-10-31
[AMD/nl] Added Dutch translation of AMD.

reinoudk:
Still a bit unsure about renaming the variables in the code examples.


adambard:
I think it'll be ok.

PR #1946, opened 2015-10-31
[haskell/en] Extended section on GHCi (discussion?)

xou:
- Added the :i command, as i feel that it is as useful as :t.
- Added another example for :t, hopefully showcasing it's flexibility
- For consistency, changed the name of (.) from function to operator
  (as is already the case with ($)), and added a short remark in the
  GHCi section that (most) operators are also functions.

### DIFF
--- a/haskell.html.markdown
+++ b/haskell.html.markdown
@@ -193,7 +193,7 @@ foo = (+10)
 foo 5 -- 15
 
 -- function composition
--- the (.) function chains functions together.
+-- the operator `.` chains functions together.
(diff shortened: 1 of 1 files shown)


haskell.html.markdown
```diff
@@ -193,7 +193,7 @@ foo = (+10)
 foo 5 -- 15
 
 -- function composition
--- the (.) function chains functions together.
+-- the operator `.` chains functions together.
```
vendethiel: That's a discutable change

vendethiel:
Thanks! :+1:

PR #1956, opened 2015-10-31
Update javascript-ua.html.markdown

Serg-Maximchuk:
Fixed translation, typos.


adambard:
Is it uk-ua or ua-ua? I think something's going to go weird if the folder name and `lang` tag don't match up (not that it's not busted now).

Serg-Maximchuk:
I looked at [cs-cz](https://github.com/adambard/learnxinyminutes-docs/blob/master/cs-cz/python3.html.markdown) - Czech, [el-gr](https://github.com/adambard/learnxinyminutes-docs/blob/master/el-gr/scala-gr.html.markdown) - Greece, and other non-symmetric lang codes, then I found all these interesting things at [ISO Language Code Table](http://www.lingoes.net/en/translator/langcode.htm), so uk-ua is correct.

Serg-Maximchuk:
I don't know how to rename folder. Can i?

Serg-Maximchuk:
Or it would be better if I will create new folder "uk-ua" and mv existing files?

Serg-Maximchuk:
Now I change folder's name to uk-ua.

PR #1957, opened 2015-10-31
[latex-de] Corrections and improvements to the german LaTeX document

jmeinke:
Fix a sentence, added a command to be able to define multiline comment blocks. Additionally, make use of the \LaTeX command to correctly show the LaTeX logo.


geoffliu:
Looks good, thanks!

PR #1959, opened 2015-10-31
[Java/ru] Update russian translation for Java.

shrz:
Note that now you can't compile both LearnJavaRu.java and LearnJava.java due to errors (multiple int fooInt declarations). I think we should fix En version and then fix Ru translation.


Oire:
Thanks!

PR #1961, opened 2015-10-31
[JSON/ms-my] Added Malay (Malaysia) translation for JSON

abdalim:


geoffliu:
Don't think any maintainer speaks Malay... We'll let public scrutiny test this instead. :) Thanks for the contribution!

PR #1966, opened 2015-10-31
Add a note regarding string value of "1"

lesaff:
Because some YAML parser will always assume that 1 is boolean for true.


Spurlow:
Thanks!

PR #1967, opened 2015-10-31
fixing javascript translate

victor-dsc:
- Removing line 439 that is already translated
- Translating line 509


Spurlow:
Thanks!

PR #1971, opened 2015-10-31
[go/fi] Translate go/en to Finnish

ComSecNinja:
I lack formal CS education in my country so some Finnish programming terms are unknown to me and I either had to (re)invent them or consult some dusty 90's programming manuals.


ComSecNinja:
The translation of the Go guide is finished.

fi-fi/go-fi.html.markdown
```diff
@@ -0,0 +1,441 @@
+---
+name: Go
```
vendethiel: that doesn't exist

fi-fi/go-fi.html.markdown
```diff
@@ -0,0 +1,441 @@
...
+category: language
+language: Go
```
vendethiel: please lowercase it

ComSecNinja:
What do you mean? This is straight from the English version. You're telling me that's wrong?

vendethiel:
Yeah, it should probably be lowercased as well, I think we lowercase all the names

ComSecNinja:
Go is written that way in English, German, French, Korean, Chinese and Russian versions at least. Those are literally every one I checked. Do you think it is wise to break the consistency here or fix all of these in one pull later?

vendethiel:
you're right, we'll fix them later

vendethiel:
Thanks!

PR #1974, opened 2015-10-31
[javascript/en] !! operator usage

zz3599:
This is useful for shorthand checking for null/falsy values.


vendethiel:
it's not null checking. it's just boolean coercion.

PR #1980, opened 2015-11-01
[CSS] Made spacing more consistent

maisyleigh:


css.html.markdown
```diff
@@ -108,7 +108,7 @@ selected:link { }
 selected:focus { }
 
 /* any element that is the first child of its parent */
-selector:first-child {}
+selector:first-child { }
 
 /* any element that is the last child of its parent */
 selector:last-child {}
```
adambard: Surely this line deserves the same treatment if the above does.

PR #1984, opened 2015-11-01
JAVA Fix grammar error

poetienshul:


hopesenddreams:
This whole section is poorly written, but I don't think your change is an improvement.

using the word maps instead of map implies that the set of Maps objects does the mapping, not the individual objects.

PR #1995, opened 2015-11-01
[latex/en] Explain how to setup a bibliography section

ramananbalakrishnan:


latex.html.markdown
```diff
@@ -227,6 +227,15 @@ format you defined in Step 1.
 
 That's all for now!
```
vendethiel: please move this line as wel

ramananbalakrishnan: This entire guide was meant to be compilable via LaTeX. Thus, the decision was based on the generated output PDF file :  [github.com/.../latex.pdf](https://github.com/ramananbalakrishnan/learnxinyminutes-docs/blob/70b3fa5bb6410b1c50e549a76807fd509119dd85/latex.pdf) 

If you scroll to the end of the document, it makes sense that the line "That's all for now!" stays above the bibliography section.

This was the reason I had chosen to place the line at this position (since "References" are usually placed after the main body text).

vendethiel:
Okay, fair enough. Thanks! :)

PR #1997, opened 2015-11-02
[Swift/zh-cn] update chinese version Swift doc

cylim:
update in Swift chinese  version.
1. deprecated println()
2. fixed getting start guide link
3. update content according to english version


vendethiel:
Thanks!

PR #1998, opened 2015-11-02
[python] Fix for unintended opposite meaning

TheRakken:
Args will be interpreted as a tuple by using *. Keyword args will be interpreted as a dictionary by using **.


vendethiel:
Thanks!

PR #2005, opened 2015-11-04
use new print instead of println

benloong:
swift 2.0 renamed println to print


benloong:
duplicated with #1997

PR #2007, opened 2015-11-04
Strings can be lexicographically compared with comparison operators

pranitbauva1997:
It is noteworthy to mention that strings can be directly compared with comparison operators which is a not available in many languages.

### DIFF
--- a/julia.html.markdown
+++ b/julia.html.markdown
@@ -102,6 +102,11 @@ false
 # Printing is easy
 println("I'm Julia. Nice to meet you!")
 
+# String can be compared lexicographically compared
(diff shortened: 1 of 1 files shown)


julia.html.markdown
```diff
@@ -102,6 +102,11 @@ false
 # Printing is easy
 println("I'm Julia. Nice to meet you!")
 
+# String can be compared lexicographically compared
```
vendethiel: two times "compared"

pranitbauva1997:
Removed the extra compared in julia.html.markdown

vendethiel:
Thanks! :+1:

PR #2021, opened 2015-11-13
[swift/en] Correct "Casting" mention in swift markdown

bobrippling:
'T()' is initialisation/creation, 'x as T' is casting


adambard:
Thanks!

PR #2028, opened 2015-11-17
Perl6: $! is not available inside the CATCH{} (Possibly)

zoffixznet:
I made this PR based on this conversation: http://irclog.perlgeek.de/perl6/2015-11-17#i_11554028 but when I tried to come up with a shorter example, I couldn't reproduce the `$!` not being set inside the `CATCH`, so now I'm unsure whether this PR is correct.

I'll leave it up to you to see whether the correction is needed.

### DIFF
--- a/perl6.html.markdown
+++ b/perl6.html.markdown
@@ -738,7 +738,7 @@ try {
 # You can throw an exception using `die`:
 die X::AdHoc.new(payload => 'Error !');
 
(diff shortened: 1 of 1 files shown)


vendethiel:
That's correct -- We use `when` to check on exception types :).

vendethiel:
Thanks again!

PR #2031, opened 2015-11-19
[erlang/en] Replaced 'or' with a semicolons

byaruhaf:
Replaced 'or' with a semicolons since the paragraph above is talking about using semicolons


Spurlow:
Thanks!

PR #2032, opened 2015-11-19
[C#/en] Grammar/clarity update in comments about File Naming

KateRita:
Clarified grammar in LearnCSharp comment about naming files relative to the classes they contain. Otherwise the two comment lines blended together and were a little confusing.


Spurlow:
Thanks!

PR #2033, opened 2015-11-19
Fix typo in Perl 6 example (closes #2025)

### LINKED ISSUE #2025: Perl 6 typo
I noticed this minor issue:

```perl6
# An array subscript can also be a closure.
# It'll be called with the length as the argument
say join(' ', @array[15..*]); #=> 15 16 17 18 19
# which is equivalent to:
say join(' ', @array[-> $n { 15..$n }]);
# Note: if you try to do either of those with an infinite loop,
#       you'll trigger an infinite loop (your program won't finish)
```

Isn't the second-to-last line supposed to say "either of those with an infinite **array**"?

kirbyfan64:
@vendethiel Done!


vendethiel:
Thanks!

PR #2035, opened 2015-11-21
[python3/cs] [python3/es] [python3/fr] [python3/ru] [python3/tr] [python3/zh] Corrects the result of math.sqrt(16)

Gnomino:
See a1ed02d6fad2b39137f52c6a04264a59e237d747


Spurlow:
Thanks!

PR #2036, opened 2015-11-23
[ruby/en] Replace Hash#has_key? with Hash#key?

fighterleslie:
1. Replace `Hash#has_key?` with `Hash#key?`
2. Replace `Hash#has_value?` with `Hash#value?`

As per the [Ruby Style Guide](https://github.com/bbatsov/ruby-style-guide#hash-key), `Hash#key?` and `Hash#value?` should be preferred to `Hash#has_key?` and `Hash#has_value?`.


vendethiel:
If a style guide says so... Thanks! :+1:

PR #2042, opened 2015-11-28
PowerShell/en: poor website render

Laoujin:
apparently ````powershell` does not render a code block on the .com website
so I've removed the 'powershell' part


vendethiel:
seems fair, thanks

PR #2049, opened 2015-12-03
[Haskell] Updating dev environment installation info

dogweather:
Current text states that Haskell is "easy" to install. Unfortunately it isn't, much of the time. I removed this boosterism. Secondly, beginners need to know that there are two commonly used options for installing Haskell; especially because the newer, Stack, produces a working dev environment more reliably.


adambard:
Wow, stack looks handy. Thanks!

PR #1164, opened 2015-07-10
[elm/en] basics of Elm in English 

GlenDC:
I've been using Elm now since about a month. And I quite like the language. as learnxinyminutes has often helped me for other languages I thought I could do something back by providing similar help to people who want to learn more about Elm. So here you can find the first version of just that.

PR #2054, opened 2015-12-08
[elm/en] Create intro to Elm

mgold:
Previously attempted at #1164 but this version is a bit more fleshed out, updated for 0.16, and I've paid special attention to the order in which concepts are introduced.

As cited in the other PR, there is no Markdown parser for Elm, so I used Haskell. It should be okay.

Feedback welcome, but the intention is to cover common cases and best practices, not be a definitive reference.

### DIFF
--- a/elm.html.markdown
+++ b/elm.html.markdown
@@ -1,0 +1,346 @@
+---
+language: Elm
+contributors:
+    - ["Max Goldstein", "http://maxgoldste.in/"]
+---
+
+Elm is a functional reactive programming language that compiles to (client-side)
+JavaScript. Elm is statically typed, meaning that the compiler catches most
+errors immediately and provides a clear and understandable error message. Elm is
+great for designing user interfaces and games for the web.
+
+
+```haskell
+-- Single line comments start with two dashes.
+{- Multiline comments can be enclosed in a block like this.
+{- They can be nested. -}
+-}
+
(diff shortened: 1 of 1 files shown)


elm.html.markdown
```diff
@@ -0,0 +1,346 @@
...
+-- Apply (call) a function by passing it arguments (no commas necessay).
+multiply 7 6 -- 42
+
+-- Partially apply a function by passing only some of its arguments.
+-- Then give that function a new name.
+double =
+  multiply 2
+
+-- Constants are similar, except there are no arguments.
+answer =
```
vendethiel: anyway reason to have a newline here and before? Reads weirdly, but maybe the conventions in Elm are different from Haskell's

mgold: That is indeed the convention, even though I don't like it, especially in this case (constants). I can ping some other Elm users if you want a third opinion.

vendethiel: No, I'll trust you :). (and I'll just blame you if someone comes later and says it's wrong... :P)

vendethiel:
Thanks!

mgold:
Thanks for merging!

The Haskell syntax highlighting works fine. The only place it really screws up is the example of invalid code (I wouldn't object to removing it). It could stand to highlight Elm's port and alias keywords, but it's not a big deal.

PR #2060, opened 2015-12-11
Updated to sdkman instead of GVM for installing groovy

ilmoralito:
sdkman (sdk in console) is the new name for gvmtool


vendethiel:
Can any other groovy-er confirm this change, please? :)

adambard:
Looks like the same tool with a new name. Works for me, thanks!

PR #2065, opened 2015-12-16
Add tutorials (Factor, Wolfram Language)

hyphz:
[factor/en]
[wolfram/en]

### DIFF
--- a/factor.html
+++ b/factor.html
@@ -1,0 +1,182 @@
+---
+language: factor
+contributors:
+    - ["hyphz", "http://github.com/hyphz/"]
+filename: learnfactor.factor
+---
+
+Factor is a modern stack-based language, based on Forth, created by Slava Pestov.
+
+Code in this file can be typed into Factor, but not directly imported because the vocabulary and import header would make the beginning thoroughly confusing.
+
+```
+! This is a comment
+
+! Like Forth, all programming is done by manipulating the stack.
+! Stating a literal value pushes it onto the stack.
+5 2 3 56 76 23 65    ! No output, but stack is printed out in interactive mode
+
+! Those numbers get added to the stack, from left to right.
+! .s prints out the stack non-destructively.
+.s     ! 5 2 3 56 76 23 65
+
+! Arithmetic works by manipulating data on the stack.
+5 4 +    ! No output
+
+! `.` pops the top result from the stack and prints it.
+.    ! 9 
+
+! More examples of arithmetic:
+6 7 * .        ! 42 
+1360 23 - .    ! 1337 
+12 12 / .      ! 1 
+13 2 mod .     ! 1 
+
+99 neg .       ! -99 
+-99 abs .      ! 99 
+52 23 max .    ! 52 
+52 23 min .    ! 23 
+
+! A number of words are provided to manipulate the stack, collectively known as shuffle words.
+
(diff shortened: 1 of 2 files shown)


factor.html
```diff
@@ -0,0 +1,182 @@
...
+13 2 mod .     ! 1 
+
+99 neg .       ! -99 
+-99 abs .      ! 99 
+52 23 max .    ! 52 
+52 23 min .    ! 23 
+
+! A number of words are provided to manipulate the stack, collectively known as shuffle words.
+
+3 dup -          ! duplicate the top item (1st now equals 2nd): 3 - 3
```
ian-bertolacci: I would suggest formatting the comments to be like Factor code, so its consistant with the way you think/write in this language:
```
3 dup -          ! duplicate the top item (1st now equals 2nd): 3 3 -
``

Comment (hyphz) [inline factor.html]: I think that might be a bit confusing for a beginner. It's meant to show the resulting arithmetic expression, not the resulting Factor code.

Comment (ian-bertolacci) [inline factor.html]: I see that point.
I know that I get confused when code simplifications dont get explained within the language.
Since this is simple arithmetic, it's fine.

File: factor.html [hunk 2]
```diff
@@ -0,0 +1,182 @@
...
+2 5 swap /       ! swap the top with the second element:        5 / 2
+4 0 drop 2 /     ! remove the top item (dont print to screen):  4 / 2
+1 2 3 nip .s     ! remove the second item (similar to drop):    1 3
+1 2 clear .s     ! wipe out the entire stack
+1 2 3 4 over .s  ! duplicate the second item to the top: 1 2 3 4 3 
+1 2 3 4 2 pick .s ! duplicate the third item to the top: 1 2 3 4 2 3
+
+! Creating Words 
+! The `:` word sets Factor into compile mode until it sees the `;` word.
+: square ( n -- n ) dup * ;    ! No output
```
vendethiel: what's `--`?

hyphz: * is multiplication, -- seperates the two halves of the stack effect. -- should be explained below but it could be clarified if necessary; that part is actually just copied from the Forth tutorial and modified for Factor's syntax changes.

vendethiel: Oh, obviously `*`, duh. Thanks.

hyphz:
Sorry about the double entries on a single pull - it seems the GitHub app won't let you put in several pull requests on a single local branch while one is outstanding.

vendethiel:
Correct - one usually just creates multiple branches, there's no risk of conflict here. What's the status on the PR?

hyphz:
I'd like to know also, I assume it's just Christmas/New Year but if this needs editing please tell me so.

vendethiel:
No, I meant it the other way. I can merge it if you're happy with what you've got here :-). But if you need some time, I can delay the merging - no rush whatsoever, of course.

hyphz:
Oh, sure, if it's acceptable then please go ahead and merge. Thanks!

vendethiel:
Thanks!

hyphz:
Thanks! I'm thinking about doing Prolog or something similar, but would it be better to do raw Prolog or a modern variant like VP or Mercury? Also, is it better to do it "imperative first" or "logic first"?

vendethiel:
I think it'd be interesting to have a logic-first Prolog tutorial, to force people to take a "different" look at it

PR #2068, opened 2015-12-17
[markdown/cz] Fix missing 'lang' in front matter

angel333:
I haven't tested this patch, but it seems that the missing parameter is causing 'markdown' to be displayed twice on the front page.

Thanks for awesome work btw!


geoffliu:
You are absolutely right. Thanks!

PR #2069, opened 2015-12-18
ColdFusion replaced &lt; and &gt; in pre block

edm00se:
The pre block was already inside a pre/code block in the rendered page, making it harder to read as they weren't being converted to the corresponding brackets (< or >).

Here's what I was seeing in the rendered page:
image


vendethiel:
Looks better, thanks!

PR #2073, opened 2015-12-24
typo

dreamtocode:
some type mistakes.
It is: syntaxtically
It should be: syntactically
It is: iLoveC
Better: ILoveC
It is: passed to ≈the function
It should be: passed to the function
It is: error
It should be: Error


vendethiel:
Thanks!

dreamtocode:
Welcome!

PR #2076, opened 2015-12-26
[perl 6] changed code output to as stated in comments

tlvb:
The comments state that Foo::Bar::inc in the package variable example should increase $Foo::Bar::n and then print it.
This did not happen, as `say ++$n;` was placed outside the block of the sub Foo::Bar::inc.
The commit puts `say ++$n;` inside the block of Foo::Bar::inc.


adambard:
Thanks!

PR #2079, opened 2015-12-27
[json/pl] Translated JSON tutorial

Voltinus:


pl-pl/json-pl.html.markdown
```diff
@@ -0,0 +1,85 @@
+---
+language: json
+filename: learnjson.json
```
vendethiel: please add `-pl` suffix to filename

vendethiel:
Thanks for the PR!

Sorry for the delay, and just as an update... Commits don't create notifications. So when you update a pull request, please comment as well so we can see it instantly :)!

PR #2080, opened 2015-12-28
[julia/en] Add different array declaration syntax

pranitbauva1997:
Add some extra type of declaration syntax for arrays


adambard:
Looks good, thanks!

PR #2096, opened 2016-01-11
[python/en] Changed a to b in comment about list assignment, and pluralized 'vari...

SWalls:
…able'.

I believe you intended to write "Point **b** to a new list". I also took the liberty of fixing a small nearby grammar mistake.


vendethiel:
Thanks!

PR #2097, opened 2016-01-11
[bash/en] fix comparison syntax

matteobaglini:


vendethiel:
Both are correct, AFAIK? Also, our guidelines file specify when to add yourself as a contributer

matteobaglini:
I removed the self addition commit. Sorry for that, I had misunderstood the contribution rules.

matteobaglini:
I tried the script while reading the guide and the `if` expression at line 86 throw an `integer expression expected` error.

According to [Advanced Bash-Scripting Guide](http://www.tldp.org/LDP/abs/html/comparison-ops.html) that error is raised because the script use integer comparison (-ne) instead of string comparison (!=).

vendethiel:
Looks good then! Thanks :). :cake:

PR #2098, opened 2016-01-13
[swift/en] Updated documentation syntax - small fix

pbrewczynski:
As Xcode switched to Markdown syntax, I've updated it.
Based on http://nshipster.com/swift-documentation/


geoffliu:
Yup, thanks for keep this up-to-date!

PR #2099, opened 2016-01-14
[Swift/en] Updated comment on documentation of Swift functions

pbrewczynski:
Documentation syntax is no longer reStructuredText,
It's "Swift-flavored version of Markdown"

http://nshipster.com/swift-documentation/


vendethiel:
Thanks!

PR #2100, opened 2016-01-16
[Swift/en] Deleted semicolon to be consistent - Small Fix

pbrewczynski:
this line 
let weak = "keyword"; let override = "another keyword" // statements can be separated by a semi-colon
show off semicolon use. But after that semicolon is not used, so I removed inconsistency.


vendethiel:
Indeed, it looks better that way :-). Thanks!

PR #2110, opened 2016-01-20
[ruby/fi] Translate Ruby docs to Finnish

firoxer:
I chose to skip translating the methods, variables and other code because they looked funny.


fi-fi/ruby-fi.html.markdown
```diff
@@ -0,0 +1,607 @@
+---
+language: ruby
+filename: learnruby.rb
```
vendethiel: please add `-fi`

adambard:
Thanks!

PR #2121, opened 2016-01-29
Update css.html.markdown

rajat-explorer:
Updated CSS for clarity.


geoffliu:
I like the edit in principle, but the wording is a bit unclear. Especially this part "rules for a block and if the rules below don't differentiate them." First off, rules don't just apply to blocks, "element" would be better. Also what does "differentiate" mean in this context? Could you change that to "rules of the same specificity" or something similar?

One more note, the contributor section up top is for people with major contributions to an article, we usually don't add people who contribute only minor patches.

rajat-explorer:
@geoffliu  Just wanted to clarify. Newcomers could be confused.

geoffliu:
Sorry it took so long, and thanks!

PR #2123, opened 2016-01-31
Update c++.html.markdown

AndrejPetelin:
v.empty() queries if v is empty while v.clear() actually clears it (or rather sets size to 0).


vendethiel:
Thanks!

PR #2127, opened 2016-02-08
[java/en] Fixed statement about many classes in one file

oscarb-se:
Fixed statement on line 441 to say that it is _not_ good practice to keep many classes in the same file.


vendethiel:
Thanks!

PR #2129, opened 2016-02-12
[zfs/en] Fix a few typos

JacobCWard:


geoffliu:
Wow, it was really broken before... Thanks!

PR #2135, opened 2016-02-15
new boolean[]  not really required when initializing.

orsenthil:
Introducing this disrupts the flow a little bit to think, if specifying `new boolean[] ` required when intializing.


geoffliu:
Yeah I think so too. :)

PR #2136, opened 2016-02-15
memoize not memorize

orsenthil:
Let's use the correct term instead of the tongue-in-cheek term denoting the same concept.


geoffliu:
Thanks!

PR #2137, opened 2016-02-15
Remove a section from c++, fixes #2130

### LINKED ISSUE #2130: [C++] "clear a container" issues
This code & comments https://github.com/adambard/learnxinyminutes-docs/commit/a230d76307ecbc0f53c4b359cdb90628720f915e are very misleading. Author mixed up few things.

1. "empty" was called instead of "clear" (but that's already fixed)
2. "Following line sets size of v to 0, but destructors don't get called," that's untrue. Instead "Following line sets size of v to 0, but capacity of vector should remain unchanged". Destructors are called, but vector will retain allocated memory.
3. "v.swap(vector<Foo>());" would reduce capacity to 0, but it isn't valid - it shouldn't compile according to C++ standard. And since C++ 11, calling "v.shrink_to_fit();" is much better choice.

After this changes, code would be correct. But is it needed? Specifics of std::vector implementation and non-common usage seem bit out of place in introduction to language.

geoffliu:


vendethiel:
we can mention the correct .clear method

geoffliu:
Eh, I do agree with the issue, that this is a bit beyond introductory guides.

vendethiel:
Okay, fair enough.

vendethiel:
ty :)

PR #2163, opened 2016-02-27
[purescript/en] comparisions -> comparisons

JacobCWard:


adambard:
Thanks for all your help with the typos!

PR #2173, opened 2016-03-04
[exlixir/en] Unclear statement on variable scope

jg-you:
In Elixir, a variable is only bound within the scope where it is declared.
In particular, a variable matched within a `case` statement will not be bound outside of the statement, see [official tutorial](http://elixir-lang.org/getting-started/case-cond-and-if.html#case).


vendethiel:
That seems good, thanks! :+1:

PR #2187, opened 2016-03-13
[haskell/de] Fixed a few typos and spelling errors

Numeri:
I changed a few simple typos (missing umlaut or missing letter, extra letter that didn't belong) and spelling errors.

### DIFF
--- a/de-de/haskell-de.html.markdown
+++ b/de-de/haskell-de.html.markdown
@@ -100,7 +100,7 @@ not False -- True
 [1..] !! 999 -- 1000
 
 -- Haskell evaluiert nun die ersten 1 - 1000 Elemente, aber der Rest der Liste
--- bleibt unangetastet. Haskell wird sie solange nicht weiterevalieren
+-- bleibt unangetastet. Haskell wird sie solange nicht weiterevaluieren
 -- bis es muss.
 
 -- Zwei Listen konkatenieren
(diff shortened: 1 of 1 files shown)


de-de/haskell-de.html.markdown
```diff
@@ -212,7 +212,7 @@ foo 5 -- 60
 -- Haskell hat einen Operator `$`, welcher Funktionsapplikation durchführt.
 -- Im Gegenzug zu der Standard-Funktionsapplikation, welche linksassoziativ ist
 -- und die höchstmögliche Priorität von "10" hat, ist der `$`-Operator
--- rechtsassoziativ und hat die Priorität 0. Dieses hat (idr.) den Effekt,
+-- rechtsassoziativ und hat die Priorität 0. Dieses hat (i.d.R.) den Effekt,
```
vendethiel: Is the capitalization intended?

Numeri: I might be wrong, but I think that's proper grammar. "i.d.R." stands for "in der Regel", (means "in general"). I think you shouldn't write "idr.", because first of all, it's three words, and second of all, the third word is capitalized if you don't abbreviate it. (I might be wrong; German isn't my native tongue)

geoffliu: My girlfriend happens to be German! "i.d.R." is the correct spelling.

PR #2201, opened 2016-03-18
Update pythonstatcomp.html.markdown

acidflip:

# Why?
I feel as if Statistical computing with Python was just "thrown" into the language section. Seeing as this is a tutorial and not an individual language, I believe it was just randomly thrown in with the rest of the languages which makes the site look a bit cluttered. Since there wasn't a category for tutorials, I think the best place to move this to is the "Tools" category.

### DIFF
--- a/pythonstatcomp.html.markdown
+++ b/pythonstatcomp.html.markdown
@@ -1,8 +1,9 @@
 ---
-language: Statistical computing with Python
+category: tool
+tool: Statistical Computing with Python
 contributors:
     - ["e99n09", "https://github.com/e99n09"]
-filename: pythonstatcomp.py
+    - ["acidflip", "https://github.com/acidflip"]
 ---
 
 This is a tutorial on how to do some typical statistical programming tasks using Python. It's intended for people basically familiar with Python and experienced at statistical programming in a language like R, Stata, SAS, SPSS, or MATLAB.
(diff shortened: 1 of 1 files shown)


acidflip:
cc @adambard

adambard:
That makes sense for now. I think I'll have to come up with some better layout one of these days and maybe sort things by language->category instead of category->language.

But, could you take your name out of the contributors? I appreciate the fix, but the page template technically assigns copyright to everyone in the contributors section, which doesn't really seem fair to the original author in this case.

acidflip:
@adambard 

> But, could you take your name out of the contributors? I appreciate the fix, but the page template technically assigns copyright to everyone in the contributors section, which doesn't really seem fair to the original author in this case.

Done! Sorry about that, I assumed anyone who edits the page is listed as a contributor there lol.

adambard:
Totally reasonable, we don't really have objective criteria or anything like we should. Thanks!

acidflip:
By the way, @adambard, how often does adambard/learnxinyminutes-docs/ sync with and update the live version of the site?

adambard:
About every 5 minutes if cron is working, and whenever I have time to check up when it's not.

acidflip:
@adambard Oh okay, I was just asking because I noticed my merge wasn't appearing on the live site but it is now.

adambard:
That has more than nothing to do with my having been reminded about it.

PR #2202, opened 2016-03-19
[kotlin/en] Add links to other resources

s-webber:
Added extra Kotlin links to the "Further Reading" section. Also removed unnecessary brackets from example of using a for loop.


vendethiel:
Thanks!

PR #2204, opened 2016-03-21
[perl6/en] Fixed instance where wrong variable name was used

williadc:
The name of the variable used is $val, but the code calls say $rand. This pull request corrects that.


adambard:
Thanks!

PR #2206, opened 2016-03-22
[c/en] issue 2205: Change variable 'size' to 'array_size'

ryebr3ad:
This is a fix for the English C page, which used two variables both named "size".  I don't anticipate this happening too often, but if someone were to copy and paste the entire file and try to compile it, an error will spring up.  To ease this burden, I changed one of the variables to "array_size", as the role of the variable was to demonstrate variable length arrays based on user input.


vendethiel:
Thanks! :star:

PR #2221, opened 2016-03-31
[latex/en] Replace invite sign $ by > in compilation example

bermanmaxim:
Using the $ sign for invite command in the `verbatim` environment messes with syntax highlighting, which stays in math mode for the rest of the document. Using > is an easy fix.


vendethiel:
Thanks

PR #2229, opened 2016-04-06
[c/en]fixed typo about boolean data types in comment in c.html.markdown

torchhound:

### DIFF
--- a/c.html.markdown
+++ b/c.html.markdown
@@ -326,7 +326,7 @@ int main (int argc, char** argv)
   using "goto" in C
   */
   typedef enum { false, true } bool;
-  // for C don't have bool as data type :(
+  // C doesn't have bool as a data type :(
   bool disaster = false;
   int i, j;
   for(i=0;i<100;++i)


c.html.markdown
```diff
@@ -326,7 +326,7 @@ int main (int argc, char** argv)
   using "goto" in C
   */
   typedef enum { false, true } bool;
-  // for C don't have bool as data type :(
+  // C doesn't have bool as a data type :(
   bool disaster = false;
```
vendethiel: This should probably be updated to actually use _Bool, or be removed

alisvlouis:
I think
> for C don't have bool as data type
and
> C doesn't have bool as a data type
are alike.

If we are talking about C99 or later instend of ANSI C, _Bool and <stdbool.h> should be mentioned.

c.html.markdown
alisvlouis: I think
> for C don't have bool as data type

and
> C doesn't have bool as a data type

are alike.


If we are talking about C99 or later instead of ANSI C, _Bool and <stdbool.h> should be mentioned.

vendethiel:
@torchhound ?

c.html.markdown
SubhrajyotiSen: The edit sounds more grammatical correct.

vendethiel:
Going to close this one for now :). Feel free to ping us here if you find some time working on it again! Thanks.

PR #2246, opened 2016-04-28
[typescript/tr-tr] TypeScript Turkish translation

zengin:
- Translated into Turkish
- Type checked in Playground to make sure that UTF-8 does not cause any problems.


vendethiel:
Thanks!

PR #2254, opened 2016-05-12
[bash/en] Fix a mistake in fgrep example

viren-nadkarni:


vendethiel:
Whoops, nice catch. Thanks.

PR #2267, opened 2016-05-27
Update javascript-es.html.markdown

imagentleman:
@vendethiel, created a new pr based on #1776, with the updates requested on your comments.

I deleted the original fork eons ago and this felt like the easiest way to address them.

Changed:

1. The conjunction, to match the original meaning (in the English article) and the context. The original word was "though", which better resembles "however" ("sin embargo" in spanish). "Aunque" is commonly used to counter something (expressed in the same sentence), closer to the way someone would use "although". The way the translation is written, it's more of a plain statement (with the conjunction at the beginning of the clause), so "Sin embargo" (however) or "Pero" (but) are better alternatives than "Aunque".
2. An Uppercase [sic] : ) typo.
3. Use of strict triple equals to match the original English article.

### DIFF
--- a/es-es/javascript-es.html.markdown
+++ b/es-es/javascript-es.html.markdown
@@ -16,7 +16,7 @@ con Java para aplicaciones más complejas. Debido a su integracion estrecha con 
 web y soporte por defecto de los navegadores modernos se ha vuelto mucho más común 
 para front-end que Java.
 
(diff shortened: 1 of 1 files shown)


vendethiel:
Thanks! Sorry again for the delay! 🚢

imagentleman:
No problem haha : )

PR #2272, opened 2016-06-04
[go/en] Updated variable references

Subzidion:
Corrected closure example that referenced assigning x to e^10, which does not happen in this iteration. Set x to a value large enough to make the comments hold true in the closure example.


adambard:
Thanks!

PR #2274, opened 2016-06-06
[tcl/en] tcl: fix error in no-braces example

Lekensteyn:
Make it exactly the same as the block before, but without braces.


vendethiel:
Good catch, thanks!

PR #2278, opened 2016-06-12
[racket/en] Add more details about Racket

g-liu:
Details added include:

* Other types of `let` expressions
* Structs
* `car` and `cdr` additional helper functions


vendethiel:
Thanks!

PR #2294, opened 2016-06-28
Remove obsolete utf-8 options.

adamheins:
These options are no longer valid in recent versions of tmux. See https://github.com/tmux/tmux/issues/230.


vendethiel:
Thanks!

PR #2307, opened 2016-07-11
[asymptotic-notation/es] add new resource link

mp19uy:
I'm modifying the "asymptotic-notation-es" file because the material from the link is in spanish.

Resource Link added: material elaborated by teachers of the course "Programación 3" (Algorithms and Data Structures) with examples about calculating the order of an algorithm and other topics in the context of asymptotic notation.
This material is publicly listed here: https://eva.fing.edu.uy/pluginfile.php/95278/mod_resource/content/0/Apuntes%20sobre%20An%C3%A1lisis%20de%20Algoritmos.pdf but I uploaded it to Scribd because I know that the link could eventually change since it's hosted in a moodle platform and now and then the teachers change the location of the files.


vendethiel:
Thanks!

PR #2338, opened 2016-08-21
c: use pointer before introduce it.

weakish:
```c
char *s = "ILoveC";
```

Here pointer `*s` is used, but the concept pointer has not been introduced above.

### DIFF
--- a/c.html.markdown
+++ b/c.html.markdown
@@ -236,7 +236,7 @@ int main (int argc, char** argv)
   z = (e > f) ? e : f; // => 10 "if e > f return e, else return f."
 
   // Increment and decrement operators:
-  char *s = "ILoveC";
+  char s[10] = "ILoveC";
   int j = 0;
   s[j++]; // => "I". Returns the j-th item of s THEN increments value of j.
   j = 0;


geoffliu:
Good point, but I do have one concern. You're changing the semantics from pointer assignment to array initialization (which is fine for this example), but we have not introduced array initialization using string literals either. So if we really want to fix this, we should add a line in the initialization section to the effect of

    char s[] = "abc";

is equivalent to

    char s[] = { 'a', 'b', 'c', '\0' };

Also, could you remove the `10`? That is certainly not the correct size, and if you initialize with string literals, the compiler will fill in the correct value for you anyway.

vendethiel:
-1. Let's not confuse people with pointers vs array anymore than they already are.

Also should be `const char *`w

geoffliu:
I was on the fence too, so let's not do this. @vendethiel

weakish:
Thanks @geoffliu, @vendethiel.

Discussing `i++` and `++i` does not need to involve pointer or array.
So I pushed a simplified version with just variable assignment.

weakish:
See #2369

PR #2339, opened 2016-08-23
added missing modulo division

zedpowa:
I noticed that the modulo division example was missing in the czech version so I added it.


vendethiel:
Thanks!

PR #2340, opened 2016-08-24
added missing js examples for czech translation

zedpowa:
I've added some more javascript examples that were missing in the czech version.


geoffliu:
Looks good (for someone who doesn't read Czech). Thanks!

PR #2350, opened 2016-08-31
[python3/en] Remove Python 2 resources

Pawtiko:
There were several resources here that teach Python 2, but not Python 3. I removed them so that a reader will only see resources that apply to Python 3.


vendethiel:
Thanks.

PR #2352, opened 2016-09-01
adding missing lang entries to headers for 3 tutorials

robochat:
These tutorials are appearing as separate entries rather than as translations.


robochat:
If you look at the webpage you'll see that there is a duplicate links for the elixir, rust and markdown tutorials, this should fix that problem.

vendethiel:
Thanks!

PR #2353, opened 2016-09-01
[python3/en] Adding "Dive Into Python 3"

Pawtiko:
The previous version of this file had the original Dive Into Python, which was written with Python 2 in mind.
It has come to my attention that the author of the original has published an updated version designed for Python 3, so I added this version back in.


vendethiel:
Thanks!

PR #2354, opened 2016-09-03
[html/en] adding translation of html/fr

robochat:
The French HTML tutorial is not visible from the webpage because there is no English language HTML tutorial. So this is an almost direct translation of the French version to English. If anything I find that this tutorial is a little basic for this site but it would fix the above problem properly.


vendethiel:
Thanks! I really messed up.

robochat:
Glad to help.

PR #2367, opened 2016-09-12
[kotlin/en] Add "if" usage as an expression

nexbit:

### DIFF
--- a/kotlin.html.markdown
+++ b/kotlin.html.markdown
@@ -262,7 +262,7 @@ fun helloWorld(val name : String) {
     } while (ctr < 10)
 
     /*
-    "if" can be used as a function that returns a value.
+    "if" can be used as an expression that returns a value.
     For this reason the ternary ?: operator is not needed in Kotlin.
     */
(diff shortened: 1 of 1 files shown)


kotlin.html.markdown
```diff
@@ -261,6 +261,14 @@ fun helloWorld(val name : String) {
         ctr++
     } while (ctr < 10)
 
+    /*
+    "if" can be used as a function that returns a value.
```
vendethiel: can be used **as an expression**

nexbit: Indeed you're right, but then I suggest the same change for the "when" keyword.. Even if it resembles a function, it's more correct to call it an expression too.
What do you think?

vendethiel: I think it's best to change `when`. Good catch.

nexbit: Ok, so I'll correct this PR and then I'll open another for `when`..

vendethiel:
Thanks

PR #2371, opened 2016-09-16
[solidity/en] corrections in mapping example

pnf408:
fixed two small errors in the example mapping called 'balances'


vendethiel:
Thanks!

PR #2405, opened 2016-10-02
[vim/en] useful tip for saving a buffer as sudo

categulario:
This could be useful


vendethiel:
why sudo?

categulario:
it is common to open a file without adequate permissions, so is good to know that there is a way to save that file in one command

vendethiel:
Thanks!

PR #2414, opened 2016-10-04
[go] added practical examples for the underscore

leonid-shevtsov:
- [x] Content changes are aimed at _intermediate to experienced programmers_ (this is a poor format for explaining fundamental programming concepts)

A coworker asked about reasons to use the underscore variable placeholder, so I decided to add some examples:
- Example of using underscore to discard the error
- Example of using underscore to loop over values of a slice
- Incidentally, example of writing to a file

### DIFF
--- a/go.html.markdown
+++ b/go.html.markdown
@@ -136,7 +136,10 @@ can include line breaks.` // Same string type.
 	// The underscore lets you "use" a variable but discard its value.
 	_, _, _, _, _, _, _, _, _, _ = str, s2, g, f, u, pi, n, a3, s4, bs
 	// Usually you use it to ignore one of the return values of a function
-	file, _ := os.Create("output.txt") // Should never ever cause an error
+	// For example, in a quick and dirty script you might ignore the
+	// error value returned from os.Create, and expect that the file
+	// will always be created.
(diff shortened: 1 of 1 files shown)


go.html.markdown
```diff
@@ -133,6 +135,11 @@ can include line breaks.` // Same string type.
 	// Unused variables are an error in Go.
 	// The underscore lets you "use" a variable but discard its value.
 	_, _, _, _, _, _, _, _, _, _ = str, s2, g, f, u, pi, n, a3, s4, bs
+	// Usually you use it to ignore one of the return values of a function
+	file, _ := os.Create("output.txt") // Should never ever cause an error
```
geoffliu: The comment here is a bit confusing. Do you mean to say file creation never causes an error in general, or that if we assume this doesn't cause an error, it would be safe to ignore the error return value with `_`?

leonid-shevtsov: Good point. I'll rephase to "we expect this call to be successful, so we're ignoring the error"

leonid-shevtsov:
@geoffliu Fixed.

vendethiel:
Thanks

PR #2425, opened 2016-10-07
Translate binary search to spanish 

ginoamaury:
Hello, I have translated file binary search to Spanish, if there is any doubt I am willing to resolve it. Thank you


vendethiel:
Thanks for both of those!

PR #2452, opened 2016-10-14
[AMD/pt_br]fix tag language pt_br for amd

diogoleal:
Fix tag language to pt_br and not ptr_br


vendethiel:
Whoops. Good catch! Thanks.

PR #2469, opened 2016-10-17
[c/pt_br]Fix correct name #define

diogoleal:
In translation for pt_br, has a mistake when written #define.


vendethiel:
Thanks!

PR #2475, opened 2016-10-18
Perl 6: Make information about Hash tables correct. Rewrite the example for dynamically scoped variables..

samcv:
Make information about Hash tables correct. Rewrite the example for dynamically scoped variables so the example is better and functions properly when ran .

Add a section on Perl 6's twigils.

Let me know what you think about the placement of the Twigils section.  It may make sense to move it or make it integrate better with the sections it relates to.


samcv:
Also, what best way to resync my repo with yours so that the old commits that have already been merged don't continue to show up?  Thanks!

perl6.html.markdown
```diff
@@ -56,28 +56,32 @@ my @array = 1, 2, 3;
 
 say @array[2]; # Array indices start at 0 -- This is the third element
 
-say "Interpolate an array using [] : @array[]";
-#=> Interpolate an array using [] : 1 2 3
+say "Access an element of an array using [] : @array[]";
```
vendethiel: No, that's wrong. I do mean interpolate. You can use "Interpolate the whole array" instead, though.

perl6.html.markdown
```diff
@@ -193,6 +198,15 @@ sub mutate($n is rw) {
...
 }
 
+my $m = 42;
+mutate $m; # $n is now 43 !
+
+# This works because we are passing the container $m to mutate.  If we try
+# to just pass a number instead of passing a variable it won't work because
+# there is no container being passed and integers are immutable by themselves:
+
+mutate 42; # Parameter '$n' expected a writable container, but got Int value
```
vendethiel: I'm not sure this commit should still be here?

perl6.html.markdown
```diff
@@ -680,11 +752,11 @@ class B is A { # inheritance uses `is`
 #  `$a .= b` is the same as `$a = $a.b`)
 # Also note that `BUILD` (the method called inside `new`)
 #  will set parent properties too, so you can pass `val => 5`.
-my B $b .= new(val => 5);
```
vendethiel: would you mind changing the variable names as well?

perl6.html.markdown
samcv: You are right I get what the intention was here.

perl6.html.markdown
samcv: I posted above: Also, what best way to resync my repo with yours so that the old commits that have already been merged don't continue to show up? Thanks!

perl6.html.markdown
samcv: Ok will do :)

samcv:
@vendethiel It should be good to merge now if you are satisfied with the new example from commit 	29bc52c

After merge I will sync my repo to yours and use branches so it doesn't always come up with things that have already been merged. :)

perl6.html.markdown
vendethiel: Sorry, didn't see it. Rebasing would work.

samcv:
That was fixed in commit 2ada588 :)

vendethiel:
Oh. Alright. WTF github, then?

Thank you very much! Sorry for being so nitpicky.

vendethiel:
🌟 :stars:

PR #2478, opened 2016-10-20
Update python3.html.markdown

morenoh149:
remove unneeded commented import

### DIFF
--- a/python3.html.markdown
+++ b/python3.html.markdown
@@ -766,7 +766,6 @@ if __name__ == '__main__':
     print(b.fly)
 
 
-# from "filename-without-extension" import "function-or-class"
 from human import Human
 from bat import Bat
 


vendethiel:
Why is that explanation unneeded?

morenoh149:
@vendethiel that module is neither from the stdlib nor is it defined within the page itself. So I can only assume it's an artifact from some previous example.

vendethiel:
It's just a syntax example.

morenoh149:
then it needs a comment explaining what's going on because I didn't get it. Moreover, if one tries to execute section 6, those imports will throw errors. They should be in a big comment block.

vendethiel:
those are fine

vendethiel:
I mean -- I'd accept a PR with more explanation. Not less.

morenoh149:
@vendethiel also note that function importing is already covered on line 621
```
# You can get specific functions from a module
from math import ceil, floor
print(ceil(3.7))   # => 4.0
print(floor(3.7))  # => 3.0
```

PR #2479, opened 2016-10-20
Add python3 class notes

morenoh149:
* explain that you can place the human and bat classes into seperate
files
* explain how to import specific functions from other files


vendethiel:
That's great! Thank you very much.

PR #2489, opened 2016-10-22
[perl/en] more perlish iterations on perl5

centreti:
- [ ] Content changes are aimed at _intermediate to experienced programmers_ (this is a poor format for explaining fundamental programming concepts)

C-style for-loops aren't Perlish, and should not be encouraged.

Also I added a `map` version, because it's worth knowing it seeing as how it's the most efficient operator for lists/arrays.

### DIFF
--- a/perl.html.markdown
+++ b/perl.html.markdown
@@ -126,17 +126,19 @@ while (condition) {
 
 
 # for loops and iteration
-for (my $i = 0; $i < $max; $i++) {
+for my $i (0 .. $max) {
   print "index is $i";
 }
 
-for (my $i = 0; $i < @elements; $i++) {
+while (my $i = each @elements) {
   print "Current element is " . $elements[$i];
 }
 
 for my $element (@elements) {
   print $element;
 }
+
+map {print} @elements;
 
 # implicitly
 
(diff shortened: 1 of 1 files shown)


perl.html.markdown
```diff
@@ -126,18 +126,20 @@ while (condition) {
 
 
 # for loops and iteration
-for (my $i = 0; $i < $max; $i++) {
+for my $i (0 .. $max) {
   print "index is $i";
 }
 
-for (my $i = 0; $i < @elements; $i++) {
+while (my $i = each @elements) {
```
vendethiel: I don't remember seeing that for an array. Is it really idiomatic? Also might warrant a version using `$#`

centreti: It isn't really idiomatic, but rather accomplishes the same thing. Honestly, I wouldn't really know of any reason to iterate over an array in this way to begin with. Might make sense just to drop this entirely.

vendethiel: Sounds about right...

Grinnz: A more idiomatic way to do this is:
`for (0..$#elements) {`

vendethiel:
Thanks!

PR #2494, opened 2016-10-22
[js / en] added a few words about js usage

DennisPS:
Added just a few words about what node.js is used for (and thereby Javascript)


vendethiel:
This is not really what people look for, imho.

PR #2505, opened 2016-10-25
[perl6] Explain `orelse` operator, add more information about exceptions, add an Iterables section...

samcv:
Make regex examples not warn and explain the warn how ambiguous spaces in regex will cause a warning.  Show how to use `m` to do regex with delimiters other than / /.  Add information about the `lazy` method  and show how you can use it with lists.  See the commit messages for full details :)

### DIFF
--- a/perl6.html.markdown
+++ b/perl6.html.markdown
@@ -1314,11 +1314,12 @@ say so 'a b c' ~~ /:s a b c /; #> `True`. We added the modifier `:s` here.
 say so 'a b c' ~~ / a b c /; #> 'False' #> Space is not significant here; please
 # use quotes or :s (:sigspace) modifier (or, to suppress this warning, omit the
 # space, or otherwise change the spacing)
-# To fix this, either use at least two spaces between strings or use the `:s`
-# adverb.
+# To fix this and make the spaces less ambiguous,  either use at least two
+# spaces between strings or use the `:s` adverb.
+
 # As we saw before, we can embed the `:s` inside the slash delimiters, but we can
 # also put it outside of them if we specify `m` for 'match':
-say so 'a b c' ~~ m:s/a b c/; #> `True`
+say so 'a b c' ~~ m:s/a  b  c/; #> `True`
 # By using `m` to specify 'match' we can also use delimiters other than slashes:
 say so 'abc' ~~ m{a  b  c}; #> `True`
 # Use the :i adverb to specify case insensitivity:
@@ -1328,37 +1329,37 @@ say so 'ABC' ~~ m:i{a  b  c}; #> `True`
 
 ## Quantifying - `?`, `+`, `*` and `**`.
 # - `?` - 0 or 1
-so 'ac' ~~ / a b c /; # `False`
-so 'ac' ~~ / a b? c /; # `True`, the "b" matched 0 times.
-so 'abc' ~~ / a b? c /; # `True`, the "b" matched 1 time.
+so 'ac' ~~ / a  b  c /; # `False`
+so 'ac' ~~ / a  b?  c /; # `True`, the "b" matched 0 times.
+so 'abc' ~~ / a  b?  c /; # `True`, the "b" matched 1 time.
 
 # ... As you read just before, whitespace is important because it determines
 #  which part of the regexp is the target of the modifier:
-so 'def' ~~ / a b c? /; # `False`. Only the `c` is optional
-so 'def' ~~ / ab?c /; # `False`. Whitespace is not significant
+so 'def' ~~ / a  b  c? /; # `False`. Only the `c` is optional
+so 'def' ~~ / a  b?  c /; # `False`. Whitespace is not significant
 so 'def' ~~ / 'abc'? /; # `True`. The whole "abc" group is optional.
 
 # Here (and below) the quantifier applies only to the `b`
 
 # - `+` - 1 or more
-so 'ac' ~~ / a b+ c /; # `False`; `+` wants at least one matching
-so 'abc' ~~ / a b+ c /; # `True`; one is enough
-so 'abbbbc' ~~ / a b+ c /; # `True`, matched 4 "b"s
+so 'ac' ~~ / a  b+  c /; # `False`; `+` wants at least one matching
+so 'abc' ~~ / a  b+  c /; # `True`; one is enough
+so 'abbbbc' ~~ / a  b+  c /; # `True`, matched 4 "b"s
 
 # - `*` - 0 or more
-so 'ac' ~~ / a b* c /; # `True`, they're all optional.
-so 'abc' ~~ / a b* c /; # `True`
-so 'abbbbc' ~~ / a b* c /; # `True`
-so 'aec' ~~ / a b* c /; # `False`. "b"(s) are optional, not replaceable.
+so 'ac' ~~ / a  b*  c /; # `True`, they're all optional.
+so 'abc' ~~ / a  b*  c /; # `True`
+so 'abbbbc' ~~ / a  b*  c /; # `True`
+so 'aec' ~~ / a  b*  c /; # `False`. "b"(s) are optional, not replaceable.
 
 # - `**` - (Unbound) Quantifier
 # If you squint hard enough, you might understand
 #  why exponentation is used for quantity.
-so 'abc' ~~ / a b ** 1 c /; # `True` (exactly one time)
-so 'abc' ~~ / a b ** 1..3 c /; # `True` (one to three times)
-so 'abbbc' ~~ / a b ** 1..3 c /; # `True`
-so 'abbbbbbc' ~~ / a b ** 1..3 c /; # `False` (too much)
-so 'abbbbbbc' ~~ / a b ** 3..* c /; # `True` (infinite ranges are okay)
+so 'abc' ~~ / a  b**1  c /; # `True` (exactly one time)
+so 'abc' ~~ / a  b**1..3  c /; # `True` (one to three times)
+so 'abbbc' ~~ / a  b**1..3  c /; # `True`
+so 'abbbbbbc' ~~ / a  b**1..3  c /; # `False` (too much)
+so 'abbbbbbc' ~~ / a  b**3..*  c /; # `True` (infinite ranges are okay)
 
 # - `<[]>` - Character classes
 # Character classes are the equivalent of PCRE's `[]` classes, but
--- a/perl6.html.markdown
+++ b/perl6.html.markdown
@@ -822,7 +822,7 @@ open 'foo' orelse say "Something happened $_"; #> Something happened
 # By using `try` and `CATCH` you can contain and handle exceptions without
 # disrupting the rest of the program. `try` will set the last exception to
 # the special variable `$!` Note: This has no relation to $!variables.
-try open 'foo'; 
+try open 'foo';
 say "Well, I tried! $!" if defined $!; #> Well, I tried! Failed to open file
                                        #foo: no such file or directory
 # Now, what if we want more control over handling the exception?
@@ -1286,7 +1286,7 @@ say @fib[^10]; #=> 1 1 2 3 5 8 13 21 34 55
 #     (grammars are actually classes)
 #  - Earliest declaration wins
 say so 'a' ~~ /a/; #=> True
-say so 'a' ~~ / a /; # More readable with some spaces!
+say so 'a' ~~ / a /; #=> True #  More readable with some spaces!
 
 # In all our examples, we're going to use the smart-matching operator against
 #  a regexp. We're converting the result using `so`, but in fact, it's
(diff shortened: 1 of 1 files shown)


perl6.html.markdown
```diff
@@ -1561,7 +1613,18 @@ for <a b c> {
...
+ is a great source of Perl 6 snippets and explainations.  If the docs don't
+ describe something well enough, you may find more detailed information here.
+ This information may be a bit older but there are many great examples and
+ explainations.  Posts stopped at the end of 2015 when the language was declared
+ stable and Perl 6.c was released.
  - Come along on `#perl6` at `irc.freenode.net`. The folks here are always helpful.
  - Check the [source of Perl 6's functions and classes](https://github.com/rakudo/rakudo/tree/nom/src/core). Rakudo is mainly written in Perl 6 (with a lot of NQP, "Not Quite Perl", a Perl 6 subset easier to implement and optimize).
  - Read [the language design documents](http://design.perl6.org). They explain P6 from an implementor point-of-view, but it's still very interesting.
+
+ [//]: # ( vim: set filetype=perl softtabstop=2 shiftwidth=2 expandtab : )
```
samcv: If you could give me what you think about having a vim modeline at the bottom to enforce the number of spaces per tabs and syntax highlighting for vim.  I have tested it and this will not show up on the page, or in the html source.  Let me know if I should remove it though.

perl6.html.markdown
```diff
@@ -1046,15 +1075,25 @@ constant thrice = gather for ^3 { say take $_ }; # Doesn't print anything
...
+say (1, 10, (20, 10) ); #> (1 10 (20 10)) Notice how grouping is maintained
+say (1, 10, (20, 10) ).flat; #> (1 10 20 10) Now the iterable is flat
 
+# - `lazy` - Defer actual evaluation until value is fetched (forces lazy context)
+my @lazy-array = (1..100).lazy;
+say @lazy-array.is-lazy; #> True # Check for lazyness with the `is-lazy` method. 
+say @lazy-array; #> [...] List has not been iterated on!
+my @lazy-array { .print }; # This works and will only do as much work as is
+# needed.
+[//]: # ( TODO explain that gather/take and map are all lazy)
```
samcv: Let me know if it's okay to have TODO comments (tested and they don't show up on the page or in the page source).

perl6.html.markdown
vendethiel: `filetype=perl6` should work.

perl6.html.markdown
vendethiel: Sounds fine. I have a lot of those in a file somewhere...

perl6.html.markdown
```diff
@@ -798,19 +798,39 @@ class Item does PrintableVal {
...
 # Exceptions are built on top of classes, in the package `X` (like `X::IO`).
-# You can access the last exception with the special variable `$!`
-# (use `$_` in a `CATCH` block) Note: This has no relation to $!variables.
+# In Perl6 exceptions are automatically 'thrown'
+open 'foo' #> Failed to open file foo: no such file or directory
+# It will also print out what line the error was thrown at and other error info
 
-# You can throw an exception using `die`:
-open 'foo' or die 'Error!'; #=> Error!
+# You can throw your own exceptions using `die`:
```
vendethiel: I think "an exception" is fine – "your own" sounds like your own `X` class

perl6.html.markdown
```diff
@@ -798,19 +798,39 @@ class Item does PrintableVal {
 
 ### Exceptions
 # Exceptions are built on top of classes, in the package `X` (like `X::IO`).
-# You can access the last exception with the special variable `$!`
-# (use `$_` in a `CATCH` block) Note: This has no relation to $!variables.
+# In Perl6 exceptions are automatically 'thrown'
+open 'foo' #> Failed to open file foo: no such file or directory
```
vendethiel: needs a semicolon

perl6.html.markdown
```diff
@@ -798,19 +798,39 @@ class Item does PrintableVal {
...
 
-# You can throw an exception using `die`:
-open 'foo' or die 'Error!'; #=> Error!
+# You can throw your own exceptions using `die`:
+die 'Error!'; #=> Error!
 # Or more explicitly:
 die X::AdHoc.new(payload => 'Error!');
 
+# In Perl 6, `orelse` is similar to the `or` operator, except it only matches
+# `undef` instead of anything evaluating as false.
```
vendethiel: Perl 6 doesn't have `undef`. "it matches exception instead of falsy values" should be correct

perl6.html.markdown
```diff
@@ -798,19 +798,39 @@ class Item does PrintableVal {
...
+# When using `orelse` it will disarm the exception and alias $_ to that failure
+# This will avoid it being automatically handled and printing lots of scary
+# error messages to the screen.
+# We can use the exception method on $_ to access the exception
+open 'foo' orelse say "Something happened {.exception}";
+# This also works, and will print out the object returned even if it's
+# not a failure:
```
vendethiel: what does that last part mean?

perl6.html.markdown
```diff
@@ -798,19 +798,39 @@ class Item does PrintableVal {
...
+# the special variable `$!` Note: This has no relation to $!variables.
+try open 'foo';
+say "Well, I tried! $!" if defined $!; #> Well, I tried! Failed to open file
+                                       #foo: no such file or directory
+# Now, what if we want more control over handling the exception?
 # Unlike many other languages, in Perl 6, you put the `CATCH` block *within*
-# the block to `try`. By default, a `try` has a `CATCH` block that catches
+# the block to `try`. Similar to how $_ was set when we 'disarmed' the
+# exception with orelse, we also use $_ in the CATCH block.
+# Note: ($! is only set *after* the `try` block)
```
vendethiel: I'm not sure $! was explained before. Maybe "$!, the last exception variable, is only set..."?

perl6.html.markdown
```diff
@@ -889,25 +909,33 @@ my $actions = JSON::Tiny::Actions.new;
 # In Perl 6, you get different behaviors based on how you declare a variable.
 # You've already seen `my` and `has`, we'll now explore the others.
 
-## * `our` (happens at `INIT` time -- see "Phasers" below)
+## * `our` declarations happen at `INIT` time -- see "Phasers" below)
```
vendethiel: extra closing paren

perl6.html.markdown
```diff
@@ -889,25 +909,33 @@ my $actions = JSON::Tiny::Actions.new;
...
 # It's like `my`, but it also creates a package variable.
 # (All packagish things (`class`, `role`, etc) are `our` by default)
-module Foo::Bar {
-  our $n = 1; # note: you can't put a type constraint on an `our` variable
-  our sub inc {
+module Var::Increment {
```
vendethiel: 👍

perl6.html.markdown
```diff
@@ -889,25 +909,33 @@ my $actions = JSON::Tiny::Actions.new;
...
+  our $our-var = 1; # Note: you can't put a type constraint like Int on an`our`
```
vendethiel: missing space before `` ` ``

perl6.html.markdown
```diff
@@ -889,25 +909,33 @@ my $actions = JSON::Tiny::Actions.new;
...
+  my $my-var = 22;  # variable.
```
vendethiel: two spaces before `#` (or is that on purpose?)

perl6.html.markdown
```diff
@@ -1046,15 +1075,25 @@ constant thrice = gather for ^3 { say take $_ }; # Doesn't print anything
...
+my @lazy-array { .print }; # This works and will only do as much work as is
+# needed.
+[//]: # ( TODO explain that gather/take and map are all lazy)
 # - `sink` - An `eager` that discards the results (forces sink context)
 constant nilthingie = sink for ^3 { .say } #=> 0 1 2
 say nilthingie.perl; #=> Nil
 
-# - `quietly` - Supresses warnings
-# Not yet implemented !
+# - `quietly` blocks will  supress warnings:
```
vendethiel: two spaces before "supress" (and should be suppress)

perl6.html.markdown
```diff
@@ -1046,15 +1075,25 @@ constant thrice = gather for ^3 { say take $_ }; # Doesn't print anything
...
+quietly { warn 'This is a warning!' }; => No output
```
vendethiel: missing `#` for comment

perl6.html.markdown
```diff
@@ -1064,7 +1103,7 @@ say nilthingie.perl; #=> Nil
 ## Everybody loves operators ! Let's get more of them
 
 # The precedence list can be found here:
-# http://perlcabal.org/syn/S03.html#Operator_precedence
+# https://docs.perl6.org/language/operators#Operator_Precedence
```
vendethiel: 👍

perl6.html.markdown
```diff
@@ -1274,50 +1313,63 @@ say so 'a' ~~ / a /; # More readable with some spaces!
...
+say so 'ABC' ~~ m:i{a  b  c}; #> `True`
 # It is, however, important as for how modifiers (that you're gonna see just below)
 #  are applied ...
 
 ## Quantifying - `?`, `+`, `*` and `**`.
 # - `?` - 0 or 1
-so 'ac' ~~ / a b c /; # `False`
-so 'ac' ~~ / a b? c /; # `True`, the "b" matched 0 times.
-so 'abc' ~~ / a b? c /; # `True`, the "b" matched 1 time.
+so 'ac' ~~ / a  b  c /; # `False`
```
vendethiel: why more space?

perl6.html.markdown
samcv: True it doesn't have 'undef' as in the case of Perl 5, but there are "undefined values" which you can check with the .defined method.  Here are the three I know of: 'Nil', 'Mu' and 'Failure'.  Maybe I should describe these types and explain them?

perl6.html.markdown
samcv: I think this could be more clear if I described Nil Mu and Failure are all 'undefined type' objects?

perl6.html.markdown
```diff
@@ -798,19 +798,39 @@ class Item does PrintableVal {
...
+open 'foo' orelse say "Something happened {.exception}";
+# This also works, and will print out the object returned even if it's
+# not a failure:
+open 'foo' orelse say "Something happened $_"; #> Something happened
+              #>  Failed to open file foo: no such file or directory
 ## Using `try` and `CATCH`
 # By using `try` and `CATCH` you can contain and handle exceptions without
-# disrupting the rest of the program.
+# disrupting the rest of the program. `try` will set the last exception to
+# the special variable `$!` Note: This has no relation to $!variables.
```
samcv: This is the first place I introduce $!  Is this part ok?  Or should rephrase as you said below?

perl6.html.markdown
samcv: That was to align it with the text above.

perl6.html.markdown
```diff
@@ -1274,50 +1313,63 @@ say so 'a' ~~ / a /; # More readable with some spaces!
...
 
 # The whitespace in a regexp is actually not significant,
-#  unless you use the `:s` (`:sigspace`, significant space) modifier.
-say so 'a b c' ~~ / a b c /; # `False`. Space is not significant here
-say so 'a b c' ~~ /:s a b c /; # `True`. We added the modifier `:s` here.
-
+#  unless you use the `:s` (`:sigspace`, significant space) adverb.
+say so 'a b c' ~~ / a  b  c /; #> `False`. Space is not significant here
+say so 'a b c' ~~ /:s a b c /; #> `True`. We added the modifier `:s` here.
+# If we use only one space between strings in a regex, Perl 6 will warn us:
```
samcv: If you do only one space Perl 6 will print out an error as I describe in this section.  So we can have two spaces or have no spaces at all.  One space between things is considered ambiguous and not best practice by Rakudo.

perl6.html.markdown
samcv: It does work, but there are sometimes glitches and it makes vim much slower moving up and down between lines than if I set the syntax highlighting to perl.

perl6.html.markdown
samcv: Also these also evaluate False with the .defined method: Any, Int when no value has been assigned to them.  There are some other types but probably aren't suited to this intro at least.

samcv: Also these also evaluate False with the .defined method: Any, Int when no value has been assigned to them.  There are some other types but probably aren't suited to this intro at least.

perl6.html.markdown
vendethiel: Seems a bit heavyweight

perl6.html.markdown
vendethiel: I think you can say "undefined values" and talk about .defined?

perl6.html.markdown
vendethiel: seems ok

samcv:
Let me know if I've resolved any of the issues you've had.  Thanks!

vendethiel:
Nice job! 🍰  Thanks for all these :)

PR #2509, opened 2016-10-26
Remove the byte order mark from the Czech Javascript file.  This file…

samcv:
… wasn't showing up on the site before but this will fix it :)

Also added how to check if the file has a BOM to the CONTRIBUTING file.


samcv:
In addition I have checked all of the docs files and this is the only one with a BOM, so all the rest should be clean. 👍

vendethiel:
Thanks :-).

PR #2514, opened 2016-10-26
Fix the xml file which was saved as ISO-8859 or Latin 1 which wasn't being generated before

samcv:
File is now in UTF-8 now.


vendethiel:
Aaah.... Thanks.

PR #2516, opened 2016-10-26
[jquery/en] Reorganisation of Part 2 ( Events and Effects )

Gnomino:
I felt like this part was kind of disorganised, with stuff being used at the beginning and explained afterwards, thus creating repetition


vendethiel:
Thanks!

PR #2523, opened 2016-10-27
[jQuery/en] Changes ambiguous class name

Gnomino:
The class name "div" might be confusing, and there was a better class name in the comments


vendethiel:
yup, much better! thank you.

PR #2531, opened 2016-10-29
c++/en typo, formatting fixes

dreamtocode:
Space should be removed.
before:
// concatenated_tuple becomes = (10, 'A', 1e9, 15, 11, 'A' ,3.14141)
after:
// concatenated_tuple becomes = (10, 'A', 1e9, 15, 11, 'A', 3.14141)
Typo initialize is correct form.
before:
map<char, int> mymap;  // Will initalize the map with key as char and value as int
after:
map<char, int> mymap;  // Will initialize the map with key as char and value as int
Typo corresponding  is correct form.
before:
// To find the value correponsing to a key
after:
// To find the value corresponding to a key


vendethiel:
Thanks!

dreamtocode:
You are welcome!

PR #2533, opened 2016-10-30
[yaml/de] improved on german translation for better understanding

Senuros:


vendethiel:
@switchhax Looks correct to you? :)
@DennisPS thanks!

switchhax:
Looks pretty good to me, but i would prefer "Erweiterung von" over "Superset", cause not everyone understands basic set theory (subset and supersets). But it is pretty optional ;)

vendethiel:
Thanks!

PR #2538, opened 2016-10-30
C/en formatting fix

dreamtocode:
Space should be removed.
Before:
  // Access struct members with .
After:
  // Access struct members with.


vendethiel:
The space is correct. The operator is dot.

dreamtocode:
My bad. You are right.

PR #2540, opened 2016-10-30
add let and const descriptions to javascript

yihongang:


vendethiel:
#1844 postponed for now :). Thanks still!

PR #2542, opened 2016-10-30
[python/en] Added note to not-named arguments .format() were added in Python 2.7

Bobowski:
Added small reminder that `"{}".foramt("Text")` works only on Python 2.7 and newer.
This might be unnecessary but for someone (like i did) writes in Python for older versions of Python might save some time in fixing errors.


vendethiel:
Not sure it brings anything; we already have a 3 tutorial.

PR #2546, opened 2016-10-31
[scala/en] Fixed confusing typo

amrue:
implicts -> implicits

Implicts is a real word synonymous with implies, but in this case the word should have been "implicits," as in implicit values. 

Also changed "favourite" -> "favorite" (US spelling) for consistency since the default language option seems to be "en-us."  


geoffliu:
Thanks!

PR #2561, opened 2016-11-10
Add support for checking YAML validity using Travis CI

samcv:
Build is passing on my branch, you can see the results here: https://travis-ci.org/samcv/learnxinyminutes-docs/

I'm not sure if @adambard would need to set this up or if somebody else can do this ( @vendethiel )?  Let me know your thoughts on this :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,3 @@
+language: ruby
+rvm:
+    - 2.2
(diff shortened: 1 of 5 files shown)


adambard:
AW YISS this is super awesome, thanks!

PR #2562, opened 2016-11-10
Update rakefile so that it will return an exit code of 1 if there is …

samcv:
Tell me what you think of this?

…a failure (Travis CI needs this). Make ISO-8859* encoding detections notices not failures

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,19 +1,23 @@
-task default: %w[test]
-
-task :test do
-  failed = false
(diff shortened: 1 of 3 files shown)


samcv:
@adambard we have a notice if it's any ISO-8859 except ISO-8859-1, but it doesn't cause a failure and it's not a real warning.  Also I was having problems getting a return code to be reliable if i didn't put it inside a test.

adambard:
Yeah, this seems less halfassed than mine.

Issue #2584, opened 2016-11-29
 [1.. 5] does not work in 0.18


anilkris:
Hi,

I liked your beginner tutorial on elm.  A small correction, I have observed.

From version 0.18,  the List need to be given as below
List.range 1 5

Please correct this, so that it would be relevant for the latest version.

Thanks,

vendethiel:
Thanks, nice catch! Would you mind sending a pull request to fix this? :-)

anilkris:
Hi,  

I have created a pull request, with the correction.  Please, let me know,  if that was ok, or if  I missed any steps in the process.

Regards,

PR #2585, opened 2016-11-29
 #2584  [1.. 5] does not work in 0.18 - elm

anilkris:
-- The second example can also be written with two dots.
List.range 1 5

-- Append lists just like strings.
List.range 1 5 ++ List.range 6 10 == List.range 1 10 -- True

-- To add one item, use "cons".
0 :: List.range 1 5  -- [0, 1, 2, 3, 4, 5]

-- The head and tail of a list are returned as a Maybe. Instead of checking
-- every value to see if it's null, you deal with missing values explicitly.
List.head ( List.range 1 5 ) -- Just 1 : Maybe.Maybe Int
List.tail ( List.range 1 5 )  -- Just [2, 3, 4, 5]

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,91 @@
+### General Guidelines
+
+Please tag your issues and pull requests with [language/lang-code] at the
+beginning (e.g. [python/en] for English Python). This will help everyone pick
+out things they care about.
+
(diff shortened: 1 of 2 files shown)


vendethiel:
Hi,

I'm not sure which steps you took to create the pull request, but it seems the two commits are old. Is your fork up-to-date?

vendethiel:
#2610

PR #2618, opened 2017-01-03
Fixed forgotten namespace std

omgee:
Fixed forgotten namespace std


vendethiel:
Why close?

remigiusz-suwalski:
@vendethiel because the namespace std wasn't forgotten, it's a good practice not to use it in large scale projects.

vendethiel:
> @vendethiel because the namespace std wasn't forgotten, it's a good practice not to use it in large scale projects.

No it's not. It's bad practice to use `using namespace std;`.

griwes:
> it's a good practice not to use it in large scale projects

Who the hell told you this nonsense?

remigiusz-suwalski:
@vendethiel @griwes [http://stackoverflow.com/questions/1452721/why-is-using-namespace-std-considered-bad-practice](These) Stack Overflow answers explain that it's a good practise **not to** use it. One should prefer `std::cout << something` over `using namespace std; cout << something`.

griwes:
@remigiusz-suwalski Ah, I see. You misunderstood the diff, and misunderstood the comments that ensued. Please do read the diff of this PR to see that it has nothing to do with `using namespace`.

ratchetfreak:
`using namespace` should **_never_** appear in global namespace. 

Inside a function scope it's fine, inside another namespace only if you are prepared to pollute the namespace.

PR #2664, opened 2017-02-17
Adds Gemfile.lock to lock charlock_holmes gem

Dorian:
Makes sure the tests are ran with a working charlock_holmes gem

e.g. prevents this from happening:

https://github.com/brianmario/charlock_holmes/issues/78

-  Pull request title is prepended with `[language/lang-code]`
- If you've changed any part of the YAML Frontmatter, make sure it is formatted according to [CONTRIBUTING.md](https://github.com/adambard/learnxinyminutes-docs/blob/master/CONTRIBUTING.markdown)
  - Yes, I have double-checked quotes and field names!

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,0 +1,15 @@
+GEM
+  remote: http://rubygems.org/
+  specs:
+    charlock_holmes (0.7.3)
+    rake (12.0.0)
+
+PLATFORMS
+  ruby
+
+DEPENDENCIES
+  charlock_holmes
+  rake
+
+BUNDLED WITH
+   1.13.7


vendethiel:
What's wrong with a basic `gem install`?

Dorian:
@vendethiel There is already a Gemfile that specify that `charlock_holmes` is needed, but the `charlock_holmes` gem doesn't work on every ruby version etc., this is a guarantee that somewhere at least someone made it work on this gem version + a ruby version (specifying a ruby version would be nice but probably a bigger change).

samcv:
What version of ruby does not support charlock_holmes? And why are we setting the version of charlock_holmes? I don't really like that.

Dorian:
bundler's documentation explains why a `Gemfile.lock` is important much better than me:

> This is important: the Gemfile.lock makes your application a single package of both your own code and the third-party code it ran the last time you know for sure that everything worked. Specifying exact versions of the third-party code you depend on in your Gemfile would not provide the same guarantee, because gems usually declare a range of versions for their dependencies.

http://bundler.io/v1.3/rationale.html

samcv:
Okay, I've decided to merge this. Thanks you for answering my questions @Dorian

PR #2676, opened 2017-02-27
[language/perl6] add much needed spacing with markdown headers

WildYorkies:

### DIFF
--- a/perl6.html.markdown
+++ b/perl6.html.markdown
@@ -25,14 +25,20 @@ double paragraphs, and single notes.
   Multiline comments use #` and a quoting construct.
   (), [], {}, 「」, etc, will work.
 )
+```
 
-### Variables
+## Variables
 
+```perl6
 # In Perl 6, you declare a lexical variable using `my`
 my $variable;
 # Perl 6 has 4 kinds of variables:
+```
 
-## * Scalars. They represent a single value. They start with a `$`
+### Scalars
+
+```perl6
+# Scalars represent a single value. They start with a `$`
 
 my $str = 'String';
 # double quotes allow for interpolation (which we'll see later):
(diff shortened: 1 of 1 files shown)


vendethiel:
Amazing, thanks. This one was really too heavy.

WildYorkies:
@vendethiel It's a great start! Still feels noisy. But I'm wondering how much of that is due to the comments not getting proper syntax highlighting. Any idea why that is?

vendethiel:
No idea :(. @samcv set us this up, but it seems [it's not highlighted correctly on her blog either](https://cry.nu/), so I'm not holding my breath for a quickfix.

WildYorkies:
I think I'll do another PR later that breaks some of the huge text blocks away from code and in normal markdown.

Like this block below should just be normal markdown. No reason it should be in the code. Unless I'm mistaken? I know there's a "download this source code" button. So maybe we need the explanation in the source?

```
## The categories are:
# - "prefix": before (like `!` in `!True`).
# - "postfix": after (like `++` in `$a++`).
# - "infix": in between (like `*` in `4 * 3`).
# - "circumfix": around (like `[`-`]` in `[1, 2]`).
# - "post-circumfix": around, after another term (like `{`-`}` in `%hash{'key'}`)
```

vendethiel:
This button downloads the whole file.

WildYorkies:
It only downloads what is inside syntax fences (from what I noticed)

samcv:
I'm guessing it may have something to do with the CSS, why the comments show up black.

WildYorkies:
@adambard Even if the comments could be greyed out _very slightly_ that could help with readability.

samcv:
This is a standard comment that isn't highlighting:
```html
<span class="c-Singleline"># Variable names can contain but not end with simple quotes and dashes,</span>
```
Looks like we need c-Singleline to have css like `cm`. 
The multi-line comments on the page are lighter and are displayed properly. Here is their html
```html
<span class="cm">  Multiline comments use #` and a quoting construct.</span>
```

Issue #2714, opened 2017-04-27
[racket/en] Incorrect name of mutable struct in setting and accessing examples.


stevechae:
In the current document, examples of setting and accessing a mutable struct called `rgba-color` are:
```rkt
(set-color-green! burgundy 10)
(color-green burgundy) ; => 10
```
This is incorrect and should be:
```rkt
(set-rgba-color-green! burgundy 10)
(rgba-color-green burgundy) ;=> 10
```

vendethiel:
Would you mind sending a PR to fix it?

stevechae:
@vendethiel sure, I'm at work now but I will submit a PR in a few hours.

PR #2715, opened 2017-04-27
[racket/en] Fix Incorrect Names in Mutable Struct Examples (per Issue #2714)

stevechae:

Per #2714 , 

Fixed incorrect name in mutable struct access and setting examples.


vendethiel:
Thank you!

Issue #2724, opened 2017-05-11
[Scala] Exampe for 'return' in anonymous function doesn't demonstrate the warned behavior


hpincket:
```
// The return keyword exists in Scala, but it only returns from the inner-most
// def that surrounds it.
// WARNING: Using return in Scala is error-prone and should be avoided.
// It has no effect on anonymous functions. For example:
def foo(x: Int): Int = {
  val anonFunc: Int => Int = { z =>
    if (z > 5)
      return z // This line makes z the return value of foo!
    else
      z + 2    // This line is the return value of anonFunc
  }
  anonFunc(x)  // This line is the return value of foo
}
```
The above example is from the current version. `foo(6)` evaluates to 6 which looks as if the anonymous function returns 6 to foo and then foo returns 6. The difference shows itself when we change the code:
```
// The return keyword exists in Scala, but it only returns from the inner-most
// def that surrounds it.
// WARNING: Using return in Scala is error-prone and should be avoided.
// It has no effect on anonymous functions. For example:
def foo(x: Int): Int = {
  val anonFunc: Int => Int = { z =>
    if (z > 5)
      return z // This line makes z the return value of foo!
    else
      z + 2    // This line is the return value of anonFunc
  }
  anonFunc(x) + 1  // This line is the return value of foo
}
```

The addition of a '+1' to the last line of foo() separates these two cases. If `return z` returns from the anonFunc then foo(6) will be 7. If `return z` returns from foo then foo(6) will be 6.

vendethiel:
True, we should use an example like yours which demonstrates the difference. Would you mind sending a PR?

cincodenada:
I came across this today myself, I've submitted a PR with the modification I did to explain it to myself, which ended up being very similar to @hpincket's suggestion.

hpincket:
Ah yes, I completely forgot about this.

PR #2769, opened 2017-06-23
[scala/en] Make return value example actually demonstrate issue

cincodenada:
Solves #2724 

Previously the `return z` didn't actually have any effect on the output, since the outer function just returned the anon function's result directly. Updated to make the outer function do something with the return value to demonstrate the difference. Also renamed functions to make what they're doing easier to follow, and added a couple examples of behavior with accompanying explanations.

### DIFF
--- a/scala.html.markdown
+++ b/scala.html.markdown
@@ -253,15 +253,19 @@ weirdSum(2, 4) // => 16
 // def that surrounds it.
 // WARNING: Using return in Scala is error-prone and should be avoided.
 // It has no effect on anonymous functions. For example:
-def foo(x: Int): Int = {
-  val anonFunc: Int => Int = { z =>
+def addTenButMaybeTwelve(x: Int): Int = {
+  val anonMaybeAddTwo: Int => Int = { z =>
     if (z > 5)
-      return z // This line makes z the return value of foo!
+      return z // This line makes z the return value of addTenButMaybeTwelve!
     else
       z + 2    // This line is the return value of anonFunc
   }
-  anonFunc(x)  // This line is the return value of foo
+  anonMaybeAddTwo(x) + 10  // This line is the return value of foo
 }
+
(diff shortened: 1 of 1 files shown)


scala.html.markdown
```diff
@@ -253,16 +253,20 @@ weirdSum(2, 4) // => 16
...
+def addTenButMaybeTwelve(x: Int): Int = {
+  val anonMaybeAddTwo: Int => Int = { z =>
     if (z > 5)
-      return z // This line makes z the return value of foo!
+      return z // This line makes z the return value of addTenButMaybeTwelve!
     else
       z + 2    // This line is the return value of anonFunc
   }
-  anonFunc(x)  // This line is the return value of foo
+  anonMaybeAddTwo(x) + 10  // This line is the return value of foo
```
vendethiel: we should probably remove that comment?

cincodenada: Certainly update with the function method, but I think it's still useful to reiterate since for many the return-last-line behavior will still be new.

Perhaps `// Last line as return value for addTenButMaybeTwelve`?

cincodenada: Or actually, it parallels nicely with the comment on z+2 above, so maybe just leave it?

vendethiel: yeah but it's not actually the return value - that'd mean the +10 gets applied everytime which isn't true.

we should maybe add "return" for the else as well

cincodenada: Well it is the return value usually, correct? It's just the explicit `return` statement above that short-circuits the function and skips the default return value. Adding `return` for the else would miss the point of the example again.

Would it be clearer to add "This line is the default return value for addTenButMaybeTwelve"? Or we could just leave the comment off and be a little less explicit, but not potentially add confusion.

vendethiel: This is the one case where the return isn't here, so I think we should not make this confusing

PR #2828, opened 2017-08-18
[angular/en] Clickable links

mikaelz:
Links are not clickable on https://learnxinyminutes.com/docs/angularjs/

ss_2017-08-18-06-24-24


adambard:
Thanks!

PR #2834, opened 2017-08-29
[ruby/en_us] Ruby has "methods" not "functions"

wwalker:
- [ww ] I solemnly swear that this is all original content of which I am the original author
- [ww ] Pull request title is prepended with `[language/lang-code]`
- [ww ] Pull request touches only one file (or a set of logically related files with similar changes made)
- [ww ] Content changes are aimed at *intermediate to experienced programmers* (this is a poor format for explaining fundamental programming concepts)
- [ww ] If you've changed any part of the YAML Frontmatter, make sure it is formatted according to [CONTRIBUTING.md](https://github.com/adambard/learnxinyminutes-docs/blob/master/CONTRIBUTING.markdown)
  - [ww ] Yes, I have double-checked quotes and field names!

### DIFF
--- a/ruby.html.markdown
+++ b/ruby.html.markdown
@@ -359,13 +359,13 @@ ensure
   puts 'This code always runs no matter what'
 end
 
-# Functions
+# Methods
 
 def double(x)
   x * 2
 end
 
-# Functions (and all blocks) implicitly return the value of the last statement
+# Methods (and all blocks) implicitly return the value of the last statement
 double(2) #=> 4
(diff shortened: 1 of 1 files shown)


duke-m:
Maybe we could keep the headline (or change it to "Subroutine") and then there could be something like "Since Ruby is thoroughly object-oriented most ruby-programmers might say rather `method` than `function`" or similar.

prertik:
Thanks!

PR #2838, opened 2017-09-05
[c++/en] container -> vector

duke-m:
fixed errors:
- "vector_name" and "Vector_name" (different case) would have resulted in a compile time error, now: "my_vector"

enhancements:
- typedef for consistency
- two push_backs to show its purpose
- both iteration types now have a working execution block (both output the vector's content)
- the first "classic loop" now also shows the operator [], which therefor is removed from below
- include and for with a white spaces for readability

### DIFF
--- a/c++.html.markdown
+++ b/c++.html.markdown
@@ -1001,13 +1001,11 @@ cout << get<5>(concatenated_tuple) << "\n"; // prints: 'A'
 // Vector (Dynamic array)
 // Allow us to Define the Array or list of objects at run time
 #include <vector>
-typedef string my_vector_type;
-my_vector_type val;
-vector<my_vector_type> my_vector; // initialize the vector
+string val;
+vector<string> my_vector; // initialize the vector
 cin >> val;
 my_vector.push_back(val); // will push the value of 'val' into vector ("array") my_vector
-cin >> val;
-my_vector.push_back(val); // will push another value into the vector
+my_vector.push_back(val); // will push the value into the vector again (now having two elements)
 
 // To iterate through a vector we have 2 choices:
 // Either classic looping (iterating through the vector from index 0 to its last index):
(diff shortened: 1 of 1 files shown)


c++.html.markdown
```diff
@@ -1000,24 +1000,26 @@ cout << get<5>(concatenated_tuple) << "\n"; // prints: 'A'
 
 // Vector (Dynamic array)
 // Allow us to Define the Array or list of objects at run time
-#include<vector>
-vector<Data_Type> Vector_name; // used to initialize the vector
+#include <vector>
+typedef string my_vector_type;
```
vendethiel: why?

duke-m: hi @vendethiel, thanks for you feedback. you mean the white space and the typedef?
I find it more useful and consistent when each subtopic is kind of cohesive. before this PR, there was "vector_name" and "Vector_name". The white space in the include has not really been questioned, I guess? :)

vendethiel: Sorry, no. I mean why the `typedef`

vendethiel: It got me confused at first because I forgot the doc has `using namespace std;` >_>

duke-m: of course we could try to have each subtopic written as a valid translatable program, here i.e. with taking care of the namespace and all includes and having a main() with a valid return, but I guess this is too fat and against the idea of learnxiny... here we could also drop the typedef because an experienced programmer will know what to do with a vector<data_type> my_vector, but I kept it to have a reader understand more quickly what this is about. at least I hoped so :)

Oire:
@vendethiel, so is it OK to be merged?

c++.html.markdown
vendethiel: I don't understand why `typedef string vector_type; vector<vector_type>` is more understandable than `vector<string>`.

duke-m:
removed the `typedef` redundancy as well as reduced the `cin` to one occurence, kept two `push_back`s. see also the commit comment

vendethiel:
Nicely done. Thank you!

PR #2856, opened 2017-09-12
[ru/all] Remove Dolphin's clutter from the repository

gim-:
This is a clutter left by [Dolphin file manager](https://userbase.kde.org/Dolphin) which has no use here.


Oire:
Thanks!

PR #2870, opened 2017-09-24
[make/zh-cn] Chinese Version of GNU Make

jcouyang:
mostly base on english version, adjust and add some examples,
also fix some targets that can't run.


prertik:
@adambard  @Menelion I think we should check and merge this. It has been here long.

prertik:
Thanks! Sorry for the delay!

PR #2871, opened 2017-09-25
[groovy/en] added missing "final" property modifier

mrosata:

### DIFF
--- a/groovy.html.markdown
+++ b/groovy.html.markdown
@@ -173,7 +173,7 @@ class Foo {
     final String name = "Roberto"
 
     // read only property with public getter and protected setter
-    String language
+    final String language
     protected void setLanguage(String language) { this.language = language }
 
     // dynamically typed property


mrosata:
I'm going to close this PR. The code suggestion isn't appropriate, I believe the term "read only" in the comment above the `language` property should have been what I removed (rather than adding a modifier to the `language` prop.

Here is the original code:
```groovy
class Foo {
    // read only property
    final String name = "Roberto"

    // read only property with public getter and protected setter
    String language
    protected void setLanguage(String language) { this.language = language }

    // dynamically typed property
    def lastName
}
```

Then here are some assertions to demonstrate that the property `language` is not "read only" in the same way `new Foo().name` is be read only (which is my understanding of read only)
```groovy

int errorCount = 0
myFoo = new Foo()

// Assert can read
assert myFoo.name == "Roberto"
try {
  // Assert can not write
  myFoo.name = "Samsonite"
} catch (e) {
  errorCount += 1
} finally {
  // The name prop should not have changed
  assert myFoo.name == "Roberto"
  assert errorCount == 1
}

// Now do the same with 'language' prop
// Assert can read
assert myFoo.language == null
try {
  // Assert can not write
  myFoo.language = "Groovy Baby"
} catch (e) {
  // bump error count when caught
  errorCount += 1
} finally {
  // The language prop should not have changed
  assert myFoo.language == null // <-- assertion fails
  assert errorCount == 2
}
```

PR #2985, opened 2017-10-29
[bf/id-id] Add Indonesian translation for brainfuck language

muhrifqii:


id-id/bf-id.html.markdown
```diff
@@ -0,0 +1,86 @@
+---
+language: "Brainfuck"
+filename: brainfuck.bf
```
prertik: lang-id is missing in the filename. The filename must be `brainfuck-id.bf`.

prertik:
Thanks!

PR #3125, opened 2018-05-12
[tr/py3] Broken OOP Example Fixed

BTaskaya:
## Broken OOP Example Fixed
-> class Insan(obje) converted class Insan(object)
-> Undefined self.name variable replaced self.isim


divayprakash:
@BTaskaya Is this ready to be merged?

PR #3216, opened 2018-09-12
Fix printf, closes #2730

### LINKED ISSUE #2730: [C/C] correction on printf
https://learnxinyminutes.com/docs/c/, these lines are wrong:

    printf("%f\n", (float)100); // %f formats a float
    printf("%lf\n", (double)100); // %lf formats a double

Yes, `%f` formats a float, but that float is actually extended to double because of varargs type promotion before the call (printf(fmt, ...) arguments after fmt are varargs). Printf function never sees floats (nor chars, nor shorts). `%lf` has no effect according to [1], expectedly.

I'd suggest the following change:

    -printf("%f\n", (float)100); // %f formats a float
    -printf("%lf\n", (double)100); // %lf formats a double
    +printf("%f\n", (double)100); // %f formats a double
    +printf("%f\n", (float)100); // floats are promoted to doubles in varargs

But the varargs topic is not covered by the article at all, so I'm unsure. Anyway, the original statement is not true (it does no harm in this context, but educates wrong things).

  [1] http://pubs.opengroup.org/onlinepubs/9699919799/functions/printf.html

>l (ell)
>Specifies that a following d, i, o, u, x, or X conversion specifier applies to a long or unsigned long argument; that a following n conversion specifier applies to a pointer to a long argument; that a following c conversion specifier applies to a wint_t argument; that a following s conversion specifier applies to a pointer to a wchar_t argument; or **has no effect** on a following a, A, e, E, f, F, g, or G conversion specifier.

divayprakash:


divayprakash:
All credits to @ian-bertolacci for the solution! I'm just carrying his work forward here to a logical conclusion. Thanks Ian!

PR #3257, opened 2018-10-05
[crystal/ru] Add Russian translation for Crystal

denpatin:


denpatin:
@divayprakash Since it's a PR, of course :)

divayprakash:
@denpatin Sorry about the wait but I'm unable to get the build to pick up this file. There seems to be something I'm missing here. Just hang on while I try to figure out what's causing the issue.

@prertik @adambard Need some help here. The build system just does not pick up this file for some reason and I cannot see why that is happening. Please take a look.

adambard:
Fixed that up, the file just had a BOM marker which the site builder can choke on.

denpatin:
@adambard It shouldn't be BOM? In the contribution guide it's written:

> You can check if the file contains a BOM on Linux/Unix systems by running `file language.html.markdown`  You will see this if it uses a BOM: `UTF-8 Unicode (with BOM) text`.

So I've saved my file as `UTF8 with BOM`...

adambard:
But prior to that:

`Try to leave out the byte-order-mark at the start of the file. (:set nobomb in Vim)`

denpatin:
Oh my gosh, I completely missed this 🤦‍♂️ Just wondering how if I've perused the whole contribution guide 😨Ok, got my fault then, thanks.

PR #3275, opened 2018-10-09
Make description of static keyword stronger, closes #2848

### LINKED ISSUE #2848: [c/en] Description of static keyword could be stronger
```
// make external variables private to source file with static:
static int j = 0; //other files using testFunc2() cannot access variable j
void testFunc2() {
  extern int j;
}
//**You may also declare functions as static to make them private**
```

It'd be perhaps better to say something like, "The static keyword makes a variable inaccessible to code outside the compilation unit. (On almost all systems, a "compilation unit" is a .c file.) static can apply both to global (to the compilation unit) variables, functions, and function-local variables. When using static with function-local variables, the variable is effectively global and retains its value across function calls, but is only accessible within the function it is declared in." Example:

```
~ $ make test
clang -Wall -Wextra -Werror -O0 -ansi -pedantic -std=c11    test.c   -o test
~ $ ./test
42 43
~ $ cat test.c
#include <stdio.h>

static int a() {
  static int i = 42;
  return i++;
}

int main() {
  printf("%d %d\n", a(), a());
}
```

You might also note that static variables are initialized to 0 if not declared with some other starting value.

divayprakash:


divayprakash:
@adambard Could you please review this?

adambard:
LGTM!

PR #3066, opened 2018-02-27
[solidity/en] Replace constant function with view function

varshard:
Replace `constant` function with `view` function as `constant` keyword in functions will be deprecated soon.

- [X] I solemnly swear that this is all original content of which I am the original author
- [X] Pull request title is prepended with `[language/lang-code]`
- [X] Pull request touches only one file (or a set of logically related files with similar changes made)
- [X] Content changes are aimed at *intermediate to experienced programmers* (this is a poor format for explaining fundamental programming concepts)
  - [X] Yes, I have double-checked quotes and field names!

PR #3340, opened 2018-10-28
Fix solidity, replace constant function with view function

divayprakash:
Cherry-picked commit 86eb5be from #3066. Author credits to @varshard for his work.


divayprakash:
@prertik Please review this before I merge it in.

prertik:
Looks good! Oh, don't forget to add the original commit authors name on the frontmatter's contributors section.

PR #3452, opened 2019-01-21
[javascript/en] Fix prototype example

sw-double:
It would also print myBoolean from prototype's prototype


divayprakash:
@Menelion Please review if possible. I'm currently having university exams so would not be able to work on this for around a week.

Menelion:
Thanks for your contribution!

PR #3524, opened 2019-04-29
[make/zh-cn] fix: typo

persevereVon:
> \# This rule will only run if file0.txt is newer than file1.txt.

Here translated into Chinese is "file0.txt 比 file1.txt 新"


inkydragon:
You may change the title to `[make/zh-cn] fix: typo`.

persevereVon:
@inkydragon OK

prertik:
Thanks for the PR!

PR #3541, opened 2019-05-31
 [rust/it] fix struct name according below usage

danielemegna:


Menelion:
@danielemegna Thanks for your contribution!

PR #3598, opened 2019-08-09
[mercury/en] Added Mercury

jrfondren:

### DIFF
--- a/mercury.html.markdown
+++ b/mercury.html.markdown
@@ -7,7 +7,7 @@ contributors:
 Mercury is a strict, pure functional/logic programming language, with
 influences from Prolog, ML, and Haskell.
 
-```mercury
+```prolog
 % Percent sign starts a one-line comment.
 
     % foo(Bar, Baz)
@@ -97,7 +97,7 @@ voter(P) :-
 
 Complete runnable example. File in 'types.m'; compile with 'mmc --make types'.
 
-```mercury
+```prolog
 :- module types.
 :- interface.
 :- import_module io.  % required for io.io types in...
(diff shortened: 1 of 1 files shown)


divayprakash:
@adambard Could you please take a look at this? The file seems fine to me, but it just throws an error when visiting ```/docs/mercury/```. The article title shows up on the homepage but even during the build ```fatal: no such path 'mercury.html.markdown' in HEAD
``` error is thrown. The browser error when visiting ```/docs/mercury/``` is -

```
Internal Server Error
Invalid argument @ io_writev - <STDERR>
WEBrick/1.4.2 (Ruby/2.5.5/2019-03-15) OpenSSL/1.1.1b at 192.168.43.67:4567
```

adambard:
@divayprakash I expect this is caused by the instruction to syntax-highlight for mercury (the  ```` ```mercury````) -- missing lexers like this will cause the page not to build.

Generally the workaround for this is to either a) pick a language for highlighting with similar-enough syntax that the highlighting generally works, b) dispense with the language specifier and have no highlighting, or c) develop a pygments highlighter for this language.

@jrfondren Do you have any input as to which of these would be most appropriate? Mercury's syntax doesn't look especially familiar to me, but perhaps there's some language with a pygments parser we could try? Failing that, perhaps we could just see what happens with a matlab parser (which I mention only as the one other language with % comments I can think of)

jrfondren:
@adambard the 'prolog' syntax is probably close enough, but if a syntax in pygments is what's needed, I don't mind trying to get Mercury into pygments first. IIRC pygments uses a bunch of regexes to highlight code, which is also what vim does, and the Mercury devs supply a vim plugin.

jrfondren:
after spending some time on this... the prolog pygment highlights Mercury just fine :)

PR #3618, opened 2019-09-01
[Rust/en] minor fixes to string slices and borrow system

Emerentius:

This PR contains 2 small fixes:
1. Explanation of string slices:
    Slices (string slices included) are not backed by two pointers but a pointer and a length. I also found this comment to be misleading: "[a slice] doesn’t actually contain the contents of a string, just a pointer to the begin[ning]". That is not a special or defining characteristic of a string slice. `String`s also just contain a pointer to the data, a length and a capacity. The real difference lies in ownership, but it's too early for that explanation.  
   Apart from the factual content, it's extraneous detail so I shortened the comment.

2. Lifetime system
    The lifetime system is no longer bound to scopes. I've updated the explanation. Several of the examples that are commented out and are supposed to cause compile errors no longer do. I've added no-op uses of the references to extend the borrows past them so that they will error again.

### DIFF
--- a/rust.html.markdown
+++ b/rust.html.markdown
@@ -92,10 +92,8 @@ fn main() {
     let s: String = "hello world".to_string();
 
     // A string slice – an immutable view into another string
-    // This is basically an immutable pair of pointers to a string – it doesn’t
-    // actually contain the contents of a string, just a pointer to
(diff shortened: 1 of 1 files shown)


divayprakash:
@Emerentius Thank you for the contribution and welcome to the LXIYM community!

PR #3763, opened 2019-11-10
[vimscript/en] Vimscript tutorial

HiPhish:

This is a tutorial for the Vim scripting language. It presumes that the reader is already familiar with Vim in general and only covers its scripting.

### DIFF
--- a/vimscript.html.markdown
+++ b/vimscript.html.markdown
@@ -40,7 +40,7 @@ pwd                   " Displays the current working directory
 
 " Except for some commands it does not; use the command delemiter before the
 " comment (echo assumes that the quotation mark begins a string)
-echo 'Hello world!'  |" Displays a message
+echo 'Hello world!'  | " Displays a message
 
 " Line breaks can be escaped by pacing a backslash as the first non-whitespace
 " character on the *following* line. Only works in script files, not on the
@@ -67,23 +67,23 @@ echo {
 " Numbers (|expr-number|)
 " #######
 
-echo  123         |" Decimal
-echo  0b1111011   |" Binary
-echo  0173        |" Octal
-echo  0x7B        |" Hexadecimal
-echo  123.0       |" Floating-point
-echo  1.23e2      |" Floating-point (scientific notation)
+echo  123         | " Decimal
+echo  0b1111011   | " Binary
+echo  0173        | " Octal
+echo  0x7B        | " Hexadecimal
+echo  123.0       | " Floating-point
+echo  1.23e2      | " Floating-point (scientific notation)
 
 " Note that an *integer* number with a leading `0` is in octal notation. The
 " usual arithmetic operations are supported.
 
-echo  1 + 2       |" Addition
-echo  1 - 2       |" Subtraction
-echo  - 1         |" Negation (unary minus)
-echo  + 1         |" Unary plus (does nothing really, but still legal)
-echo  1 * 2       |" Multiplication
-echo  1 / 2       |" Division
-echo  1 % 2       |" Modulo (remainder)
+echo  1 + 2       | " Addition
+echo  1 - 2       | " Subtraction
+echo  - 1         | " Negation (unary minus)
+echo  + 1         | " Unary plus (does nothing really, but still legal)
+echo  1 * 2       | " Multiplication
+echo  1 / 2       | " Division
+echo  1 % 2       | " Modulo (remainder)
 
 " Booleans (|Boolean|)
 " ########
@@ -92,19 +92,19 @@ echo  1 % 2       |" Modulo (remainder)
 " converted to numbers (see below). There are two pre-defined semantic
 " constants.
 
-echo  v:true      |" Evaluates to 1 or the string 'v:true'
-echo  v:false     |" Evaluates to 0 or the string 'v:false'
+echo  v:true      | " Evaluates to 1 or the string 'v:true'
+echo  v:false     | " Evaluates to 0 or the string 'v:false'
 
 " Boolean values can result from comparison of two objects.
 
-echo  x == y             |" Equality by value
-echo  x != y             |" Unequality
-echo  x >  y             |" Greater than
-echo  x >= y             |" Greater than or equal
-echo  x <  y             |" Smaller than
-echo  x <= y             |" Smaller than or equal
-echo  x is y             |" Instance identity (lists and dictionaries)
-echo  x isnot y          |" Instance non-identity (lists and dictionaries)
+echo  x == y             | " Equality by value
+echo  x != y             | " Unequality
+echo  x >  y             | " Greater than
+echo  x >= y             | " Greater than or equal
+echo  x <  y             | " Smaller than
+echo  x <= y             | " Smaller than or equal
+echo  x is y             | " Instance identity (lists and dictionaries)
+echo  x isnot y          | " Instance non-identity (lists and dictionaries)
 
 " Strings are compared based on their alphanumerical ordering
 " echo 'a' < 'b'. Case sensitivity depends on the setting of 'ignorecase'
@@ -118,19 +118,19 @@ echo  'a' <? 'B'         | " True
 echo  'a' <# 'B'         | " False
 
 " Regular expression matching
-echo  "hi" =~  "hello"    |" Regular expression match, uses 'ignorecase'
-echo  "hi" =~# "hello"    |" Regular expression match, case sensitive
-echo  "hi" =~? "hello"    |" Regular expression match, case insensitive
-echo  "hi" !~  "hello"    |" Regular expression unmatch, use 'ignorecase'
-echo  "hi" !~# "hello"    |" Regular expression unmatch, case sensitive
-echo  "hi" !~? "hello"    |" Regular expression unmatch, case insensitive
+echo  "hi" =~  "hello"    | " Regular expression match, uses 'ignorecase'
+echo  "hi" =~# "hello"    | " Regular expression match, case sensitive
+echo  "hi" =~? "hello"    | " Regular expression match, case insensitive
+echo  "hi" !~  "hello"    | " Regular expression unmatch, use 'ignorecase'
+echo  "hi" !~# "hello"    | " Regular expression unmatch, case sensitive
+echo  "hi" !~? "hello"    | " Regular expression unmatch, case insensitive
 
 " Boolean operations are possible.
 
-echo  v:true && v:false       |" Logical AND
-echo  v:true || v:false       |" Logical OR
-echo  ! v:true                |" Logical NOT
-echo  v:true ? 'yes' : 'no'   |" Ternary operator
+echo  v:true && v:false       | " Logical AND
+echo  v:true || v:false       | " Logical OR
+echo  ! v:true                | " Logical NOT
+echo  v:true ? 'yes' : 'no'   | " Ternary operator
 
 
 " Strings (|String|)
@@ -140,9 +140,9 @@ echo  v:true ? 'yes' : 'no'   |" Ternary operator
 " depends on the option |'encoding'|.
 
 " Literal constructors
-echo  "Hello world\n"   |" The last two characters stand for newline
-echo  'Hello world\n'   |" The last two characters are literal
-echo  'Let''s go!'      |" Two single quotes become one quote character
+echo  "Hello world\n"   | " The last two characters stand for newline
+echo  'Hello world\n'   | " The last two characters are literal
+echo  'Let''s go!'      | " Two single quotes become one quote character
 
 " Single-quote strings take all characters are literal, except two single
 " quotes, which are taken to be a single quote in the string itself. See 
@@ -150,21 +150,21 @@ echo  'Let''s go!'      |" Two single quotes become one quote character
 
 " String concatenation
 " The .. operator is preferred, but only supported in since Vim 8.1.1114
-echo  'Hello ' .  'world'  |" String concatenation
-echo  'Hello ' .. 'world'  |" String concatenation (new variant)
+echo  'Hello ' .  'world'  | " String concatenation
+echo  'Hello ' .. 'world'  | " String concatenation (new variant)
 
 " String indexing
-echo  'Hello'[0]           |" First byte
-echo  'Hello'[1]           |" Second byte
-echo  'Hellö'[4]           |" Returns a byte, not the character 'ö'
+echo  'Hello'[0]           | " First byte
+echo  'Hello'[1]           | " Second byte
+echo  'Hellö'[4]           | " Returns a byte, not the character 'ö'
 
 " Substrings (second index is inclusive)
-echo  'Hello'[:]           |" Copy of entire string
-echo  'Hello'[1:3]         |" Substring, second to fourth byte
-echo  'Hello'[1:-2]        |" Substring until second to last byte
-echo  'Hello'[1:]          |" Substring with starting index
-echo  'Hello'[:2]          |" Substring with ending index
-echo  'Hello'[-2:]         |" Substring relative to end of string
+echo  'Hello'[:]           | " Copy of entire string
+echo  'Hello'[1:3]         | " Substring, second to fourth byte
+echo  'Hello'[1:-2]        | " Substring until second to last byte
+echo  'Hello'[1:]          | " Substring with starting index
+echo  'Hello'[:2]          | " Substring with ending index
+echo  'Hello'[-2:]         | " Substring relative to end of string
 
 " A negative index is relative to the end of the string. See
 " |string-functions| for all string-related functions.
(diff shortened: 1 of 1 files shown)


vimscript.html.markdown
```diff
@@ -0,0 +1,572 @@
...
+```
+
+### Executing (run-time macros of sorts)
+
+Sometimes we need to construct an ex-command where part of the command is not
+known until runtime.
+
+```vim
+let line = 3               |" Line number deternimed at runtime
+execute line . 'delete'    |" Delete a line
```
karismatic-megafauna: does this delete line 3? not really clear to me what the `.` is doing and if "delete" is a keyword or something.

vimscript.html.markdown
```diff
@@ -0,0 +1,572 @@
...
+#### Auto groups
+
+When a file is source multiple times the auto-commands are defined anew,
+without deleting the old ones, causing auto-commands to pile up over time. Use
+auto-groups and the following ritual to guard against this.
+
+```vim
+augroup sgml-like   |" The name of the group is arbitrary
+		autocmd!        |" Deletes all auto-commands in the current group
+		autocmd Filetype html,xml :set tabstop=2
```
dhruvasagar: Should use `:setlocal` here

vimscript.html.markdown
```diff
@@ -0,0 +1,572 @@
...
+echo  {'x': {'a': 1, 'b': 2}}  |" Nested dictionary
+
+" Indexing a dictionary
+echo  {'a': 1, 'b': 2}['a']    |" Literal index
+echo  {'a': 1, 'b': 2}.a       |" Syntactic sugar for simple keys
+```
+
+See `:h dict-functions` for dictionary manipulation functions.
+
+### Funcref
```
RRethy: Might be worth adding that a funcref must be stored in a variable that begins with a capital letter.

vimscript.html.markdown
```diff
@@ -0,0 +1,572 @@
...
+    function! Adder(n) closure
+        return n + x
+    endfunction
+    return funcref('Adder')
+endfunction
+let AddFive = MakeAdder(5)
+echo AddFive(3)               |" Prints 8
+
+" Aborting functions, abort once error occurs (:h :func-abort)
+function! SourceMyFile() abort
```
RRethy: Maybe add comment stating the `abort` was added after the function name since a reader may miss this.

vimscript.html.markdown
```diff
@@ -0,0 +1,572 @@
...
+See `:h 40.2`, `:h :command`
+
+```vim
+" First argument is the name of the command, rest is the command body
+command! SwapLines normal ddp
+```
+
+The exclamation mark works the same as with `:function`. User-defined commands
+must start with a capital letter. The `:command` command can take many extra
+optional arguments, such as `-nargs`, all of them start with a dash.
```
RRethy: [Here](https://github.com/adambard/learnxinyminutes-docs/pull/3763/files#diff-eca009cb1a774d802a20c34b44b8e723R472) you said the first argument is the command name. This comment sort of contradicts this since if we call `-nargs` an optional argument, then the first argument might not be the command name. Although, it might be better to change this from `optional arguments` to `options` and not change the previous comment.

vimscript.html.markdown
```diff
@@ -0,0 +1,572 @@
...
+" Source first match among runtime paths
+runtime plugin/my-plugin.vim
+```
+
+### Defining new ex-commands
+See `:h 40.2`, `:h :command`
+
+```vim
+" First argument is the name of the command, rest is the command body
+command! SwapLines normal ddp
```
RRethy: NIT: This command name would insinuate that it works on a range. If you just want a simple command to illustrate the comment then something along the lines of

```vim
command! DeleteLine normal! dd
```

might be better.

Also, to keep with best practices and what you mention below, should probably add the bang onto `:normal`.

vimscript.html.markdown
```diff
@@ -0,0 +1,572 @@
...
+
+The exclamation mark works the same as with `:function`. User-defined commands
+must start with a capital letter. The `:command` command can take many extra
+optional arguments, such as `-nargs`, all of them start with a dash.
+
+### Defining auto-commands
+See `:h 40.3`, `:h autocmd`
+
+```vim
+" Events, patterns, commands
```
RRethy: NIT: confusing comment IMO.

vimscript.html.markdown
RRethy: @karismatic-megafauna Just an FYI, this string concatenation was mentioned [here](https://github.com/adambard/learnxinyminutes-docs/pull/3763/files#diff-eca009cb1a774d802a20c34b44b8e723R122).

vimscript.html.markdown
```diff
@@ -0,0 +1,572 @@
...
+
+
+" Test if something exists
+echo  exists('g:my_var')           |" Variable (scope optional)
+echo  exists('MyFunc')             |" Function
+echo  exists('my_dict["foo"]')     |" Dictionary entry
+echo  exists('my_dict["foo"]')     |" Dictionary entry
+echo  exists('$TERM')              |" Environment variable
+echo  exists(':echo')              |" Command
+echo  exists('+tui')               |" Feature
```
RRethy: Adding more examples here would be beneficial. This set of examples from `:h exists()` is more comprehensive. (Also line 54[12] are duplicates).

```vim
echo exists("&mouse")                     |" Option exists
echo exists("+mouse")                     |" Option exists && works
echo exists("$HOSTNAME")                  |" Environment variable
echo exists("*strftime")                  |" Built-in function
echo exists("*s:MyFunc")                  |" User defined function
echo exists("bufcount")                   |" Variable
echo exists(":Make")                      |" Command
echo exists("#CursorHold")                |" autocmd for this event in the current augroup
echo exists("#BufReadPre#*.gz")           |" autocmd for this event and pattern
echo exists("#filetypeindent")            |" augroup
echo exists("#filetypeindent#FileType")   |" augroup and autocmd for this event
echo exists("#filetypeindent#FileType#*") |" augroup and autocmd for this event and pattern
echo exists("##ColorScheme")              |" autocmd for this event is supported
```

RRethy: Also, line 540 isn't entirely accurate.

vimscript.html.markdown
```diff
@@ -0,0 +1,572 @@
...
+
+" Various dynamic values
+echo  expand('%')                  |" Current file name
+echo  expand('<cword>')            |" Current word under cursor
+echo  expand('%:p')                |" Modifier, get full path
```
RRethy: Adding another section just for modifiers would be useful IMO.

vimscript.html.markdown
```diff
@@ -0,0 +1,572 @@
...
+
+```
+" String concatenation
+echo  'Hello ' .  'world'  |" String concatenation
+echo  'Hello ' .. 'world'  |" String concatenation
+
+" String indexing
+echo  'Hello'[0]           |" First byte
+echo  'Hello'[1]           |" Second byte
+echo  'Hello'[-1]          |" Last byte
```
mg979: This doesn't work for me, `'hello'[-1:-1]` works, can you test it in vim?

vimscript.html.markdown
```diff
@@ -0,0 +1,572 @@
...
+" Substrings
+echo  'Hello'[:]           |" Copy of entire string
+echo  'Hello'[1:3]         |" Substring, second to fourth byte
+echo  'Hello'[1:-2]        |" Substring until second to last byte
+echo  'Hello'[1:]          |" Substring with starting index
+echo  'Hello'[:2]          |" Substring with ending index
+```
+
+A negative index is relative to the end of the string. See `:h
+string-functions` for all string-related functions.
```
mg979: You could point out that, differently from python, the second index in the slice is *included*, so that `hello[:-1]` is unfortunately the same as `hello[:]` and `hello`, and if you want the string *up to but not including the last byte* you should write `hello[:-2]`, for the same reason line 135 could be wrong (it's a substring *until the last byte*, not the second to last byte)

vimscript.html.markdown
mg979: When an option is only local, it doesn't matter (tabstop is always local)

vimscript.html.markdown
```diff
@@ -0,0 +1,572 @@
...
+contributors:
+    - ["HiPhish", "http://hiphish.github.io/"]
+---
+
+## Introduction
+
+Here is a dirty secret: there is no such thing as a Vim scripting language.
+There are ex-commands, a subset of which implements features one would expect
+from a scripting language (variable bindings, functions, loops). This is what
+people usually mean when they speak of Vim script (also called VimL). Always
```
mg979: My opinion, but in this sense vimL isn't less of a scripting language that bash is, and bash is considered a scripting language. Not to mention python and others have their interpreteres so how is it that they differ? The difference is just that the interpreter for vimL is vim itself, but it's still a scripting language. It even supports `self` for dictionary functions, lambdas etc making it much more complex than bash (a scripting language). Maybe you should say that vim Ex commands are integral part of the language, just as regular executable commands can be used in bash scripts, I think it would be more fitting.

vimscript.html.markdown
```diff
@@ -0,0 +1,572 @@
...
+```
+
+Single-quote strings take all characters are literal, except two single quotes,
+which are taken to be a single quote in the string itself. See `:h expr-quote`
+for all possible escape sequences.
+
+```
+" String concatenation
+echo  'Hello ' .  'world'  |" String concatenation
+echo  'Hello ' .. 'world'  |" String concatenation
```
kevinushey: Is there any difference between `.` and `..`?

vimscript.html.markdown
```diff
@@ -0,0 +1,572 @@
...
+endtry
+```
+
+
+## Functions
+
+### Defining functions
+
+```vim
+" Function names have to start with a capital letter
```
kevinushey:
```suggestion
" Unscoped function names have to start with a capital letter
```

kevinushey: Perhaps unnecessary since this is made clear below with `s:addNumersLoudly(x, y)`?

vimscript.html.markdown
```diff
@@ -0,0 +1,572 @@
...
+echo  type(my_var) == type(function('type'))   |" Funcref
+```
+
+
+## Tricks of the trade
+
+### Source guard
+
+```vim
+" Prevent a file from being source multiple times; users can set the variable
```
kevinushey:
```suggestion
" Prevent a file from being sourced multiple times; users can set the variable
```

vimscript.html.markdown
```diff
@@ -0,0 +1,572 @@
...
+See `:h Dictionary`
+
+An unordered sequence of key-value pairs, keys are always strings (numbers are
+implicitly converted to strings).
+
+```vim
+" Dictionary literal
+echo  {}                       |" Empty dictionary
+echo  {'a': 1, 'b': 2}         |" Dictionary literal
+echo  {'x': {'a': 1, 'b': 2}}  |" Nested dictionary
```
crescentrose: I'd add that you can make multiline dictionaries by prefixing new lines with a backslash:

```vim
let foo = {
  \ 'key': 'value',
  \ 'another_key': 'another_value'
\}
```

This is not immediately obvious but is very helpful for code readability as well as used frequently in plugin configuration.

HiPhish:
> In the section on scope and dictionaries, it might be good to add access using `get` for default values. Such as `get(b:, 'varname', g:varname)` and `get(a:, 1, :default')`. The latter could benefit from an explanation using optional arguments.

I would rather put that near the end among the tricks of the trade. I have already mentioned that scopes are dictionaries, so that would be just bloat up there.

> Additionally, in the regex section, `=~` behavior depends on case settings, so using `=~#` or `=~?` is recommended.

Good point.

vimscript.html.markdown
```diff
@@ -0,0 +1,572 @@
...
+endfor
+
+" Iterate over the keys of a dictionary
+for symbol in keys({'π': 3.14, 'e': 2.71})
+		echo 'The constant ' . symbol . ' is a transcendent number'
+endfor
+
+" Iterate over the values of a dictionary
+for value in keys({'π': 3.14, 'e': 2.71})
+		echo 'The value ' . symbol . ' approximates a transcendent number'
```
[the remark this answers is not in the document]
HiPhish: Nice catch, but you missed the mistake in the line above ;) I should be iterating over the values instead of the keys.

vimscript.html.markdown
dhruvasagar: @mg979 It's better to be specific and understand the implications.

vimscript.html.markdown
```diff
@@ -0,0 +1,572 @@
...
+```
+
+
+## Functions
+
+### Defining functions
+
+```vim
+" Function names have to start with a capital letter
+function! AddNumbersLoudly(x, y)
```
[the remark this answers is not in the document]
HiPhish: It echoes a message and then returns the result. I wanted to have a function with both a side effect and a return value.

vimscript.html.markdown
```diff
@@ -0,0 +1,572 @@
...
+" Can start with lower-case letter if local to script
+function! s:addNumbersLoudly(x, y)
+    echo 'Adding'  . a:x .  'and'  . a:y
+    return a:x + a:y
+endfunction
+
+" Closures, functions carrying values from outer scope (:h :func-closure)
+function! MakeAdder(x)
+    function! Adder(n) closure
+        return n + x
```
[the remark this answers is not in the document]
HiPhish: You are absolutely right, the scope was missing.

vimscript.html.markdown
```diff
@@ -0,0 +1,572 @@
...
+```vim
+" Call a function for its return value, and possibly for its side effects
+let animals = keys({'cow': 'moo', 'dog': 'woof', 'cat': 'meow'})
+
+" Call a function for its side effects only, ignore potential result
+call sign_undefine()
+```
+
+Recall that Vim script is embedded within the ex-commands, that is why we
+cannot just call a function directly, we have to use the `:call` ex-command.
```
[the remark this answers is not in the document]
HiPhish: Good point.

vimscript.html.markdown
HiPhish: Good point.

vimscript.html.markdown
HiPhish: I have split the code block and added a sentence for all the special annotations. That should make it clear.

vimscript.html.markdown
HiPhish: I clarified it a bit.

vimscript.html.markdown
HiPhish: Fixed.

vimscript.html.markdown
HiPhish: Good point.

vimscript.html.markdown
HiPhish: Fixed.

HiPhish:
> Small comments for the most part. I think adding more _best practices_ would be beneficial. Also, I'm not sure I see the purpose of always using `|"` to start a line comment when `"` should suffice.

That won't work with `:echo`, it will assume that the quote is opening a new string. Using `|` to delimit the `:echo` works though.

vimscript.html.markdown
HiPhish: Right, that was a mistake.

vimscript.html.markdown
HiPhish: I added a little note that the second index is inclusive. It is worth mentioning since in most languages the second index is exclusive, but I agree that comparisons with other languages are out of place here.

vimscript.html.markdown
HiPhish: You are right, I think that was a bit too harsh.

vimscript.html.markdown
HiPhish: I have added a note.

vimscript.html.markdown
HiPhish: Good point.

vimscript.html.markdown
```diff
@@ -0,0 +1,572 @@
...
+### Defining auto-commands
+See `:h 40.3`, `:h autocmd`
+
+```vim
+" Events, patterns, commands
+autocmd Filetype html,xml :set tabstop=2
+```
+
+Events and patterns are separated by commas with no space between. See `:h
+autocmd-events` for standard events.
```
dhruvasagar: I think it's important to mention autocmd event User

vimscript.html.markdown
HiPhish: Good point. It should then also be mentioned for multi-line lists.

vimscript.html.markdown
```diff
@@ -0,0 +1,572 @@
...
+
+```vim
+let b:my_var = 1        |" Local to current buffer
+let w:my_var = 1        |" Local to current window
+let t:my_var = 1        |" Local to current tab page
+let g:my_var = 1        |" Global variable
+let l:my_var = 1        |" Local to current function (see functions below)
+let s:my_var = 1        |" Local to current script file
+
+" The following scopes are read-only
```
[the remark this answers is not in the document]
HiPhish: Fixed.

vimscript.html.markdown
```diff
@@ -0,0 +1,572 @@
...
+
+" Access scopes as dictionaries (can be modified like all dictionaries)
+echo  b:                |" All buffer variables
+echo  w:                |" All window variables
+echo  t:                |" All tab page variables
+echo  g:                |" All global variables
+echo  l:                |" All local variables
+echo  s:                |" All script variables
+echo  a:                |" All function arguments
+echo  v:                |" All Vim variables
```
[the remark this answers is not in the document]
HiPhish: Sure.

vimscript.html.markdown
```diff
@@ -0,0 +1,572 @@
...
+without deleting the old ones, causing auto-commands to pile up over time. Use
+auto-groups and the following ritual to guard against this.
+
+```vim
+augroup sgml-like   |" The name of the group is arbitrary
+		autocmd!        |" Deletes all auto-commands in the current group
+		autocmd Filetype html,xml :set tabstop=2
+augroup END         |" Switch back to default auto-group
+```
+
```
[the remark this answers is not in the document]
HiPhish: Good point.

vimscript.html.markdown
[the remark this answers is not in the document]
HiPhish: Good point.

HiPhish:
Thank you all for your input, I have incorporated most of the feedback :)

vimscript.html.markdown
mg979: ok but vimL is the only language I ever saw with this (questionable) characteristic, I think it's important to point it out.

vimscript.html.markdown
mg979: This should go together with command concatenation `|`, line continuations `\` are a generic syntactic element, not limited to lists, dicts, etc, you can use them for commands, `if` statements etc.

mg979:
> 
> 
> > In the section on scope and dictionaries, it might be good to add access using `get` for default values. Such as `get(b:, 'varname', g:varname)` and `get(a:, 1, :default')`. The latter could benefit from an explanation using optional arguments.
> 
> I would rather put that near the end among the tricks of the trade. I have already mentioned that scopes are dictionaries, so that would be just bloat up there.
> 

I think a section with some commonly used built-in functions would be useful
```vim
get(dict, key, default)         " get value from dictionary, return default if key is absent
printf('%-20s', left_aligned)   " format a string
expand('<cword>')               " expand vim-specific patterns, or paths
" and more... 
```

vimscript.html.markdown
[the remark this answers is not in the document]
karismatic-megafauna: I see, in my mind loudly is LIKE THIS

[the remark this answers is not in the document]
karismatic-megafauna: that is to say that loudly would convert to all caps

vimscript.html.markdown
karismatic-megafauna: I am not sure it was mentioned at the time of me commenting, I think the author rebased as my comments are out of date and their is still one commit... :/ 

Thank you for pointing it out tho!

HiPhish:
*bump*

Anything else to be done before this can be merged?

divayprakash:
@HiPhish Thank you for the contribution and a big welcome to the LXIYM community!

Sorry about the delay in merging, I'm working on this now. Should be done by the end of next week.

divayprakash:
@HiPhish Thanks again for all your work on this!

PR #3825, opened 2020-01-24
[smalltalk/en] Made some improvements

David-Hari:
• Under Simple syntax, it referred to a unary message as “a single string”. I think it would be clearer and more accurate to describe it as “a single symbol”.  Calling it a string may leave people confused with actual strings (i.e. instance of String class). Furthermore, message selectors are actually Symbol instances.
• Fixed the formatting of the bullet-point list of messages.
• Fixed the formatting of the example code, and set highlighting to smalltalk language.
• Under Assignment, many of the examples are actually access and assignment. You could simplify the list down to just the first 6 (i.e. up to and including x := Object new). The rest can go in a section further down which is about interesting base access methods. Even the Object new example isn’t really necessary, but it helps to show that assigning objects is no different from assigning literals.

### DIFF
--- a/smalltalk.html.markdown
+++ b/smalltalk.html.markdown
@@ -30,7 +30,7 @@ The most basic operation is to send a message to an object
 `anObject aMessage`
 
 There are three sorts of messages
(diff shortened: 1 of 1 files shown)


adambard:
Thanks! I'm not 100% that the smalltalk lexer exists, but let's try it out! (if you discover that smalltalk went away, you know where to find me)

Issue #3592, opened 2019-08-04
[C/de] adding german language


LamdaLamdaLamda:
Translating the english c version to german.

caminsha:
Started to translate C to German.
You can see progress [here](https://github.com/caminsha/learnxinyminutes-docs/blob/de-de/c/de-de/c-de.html.markdown).

Maybe I will need some days for it as it is a quite long article.

divayprakash:
WIP #3839

mxschumacher:
The translation has been merged, thanks @caminsha !

PR #3839, opened 2020-02-01
[de-de/c] Translated C to German (See Issue #3592)

caminsha:
I have translated the English version of C to German.
This PR will solve Issue #3592.

Sometimes it was quite difficult to translate the words because the content would make less sense when translating to German. I hope it's not that bad.

### DIFF
--- a/de-de/c-de.html.markdown
+++ b/de-de/c-de.html.markdown
@@ -865,7 +865,4 @@ es wichtig, dass der Codestil möglichst konsistent ist. Es ist wichtiger, lesba
 Code zu schreiben als Code, welcher clever und schnell ist. Es lohnt sich ein
 Blick auf den [Codestil des Linuxkernel](https://www.kernel.org/doc/Documentation/process/coding-style.rst) zu werfen. [Englisch]
 
-Wenn die erwähnte Literatur dein Problem nicht löst, denke daran:
-"Google ist dein Freund"
-
 [1] [Why isn't sizeof for a struct equal to the sum of sizeof of each member?](http://stackoverflow.com/questions/119123/why-isnt-sizeof-for-a-struct-equal-to-the-sum-of-sizeof-of-each-member)
--- a/de-de/c-de.html.markdown
+++ b/de-de/c-de.html.markdown
@@ -16,7 +16,7 @@ manuellen Speicherverwaltung bewusst sein.
 
 > **Über Compiler Optionen**
 > 
-> Standardmässig sind `gcc` und `clang` ziemlich ruhig bezüglich Warnungen und 
+> Standardmäßig sind `gcc` und `clang` ziemlich ruhig bezüglich Warnungen und 
 > Fehlern, obwohl dies sehr nützliche Informationen sein können. Es wird
 > empfohlen, strengere Compiler Optionen zu verwenden. Hier sind einige empfohlene
 > Standards:
@@ -203,7 +203,7 @@ int main (int argc, char** argv){
     double x_double = 0.0; // echte Zahlen ohne Suffix sind vom Typ double
 
     // integer-Typen können vorzeichenlos (unsigned) sein
-    // (grösser oder kleiner als 0)
+    // (größer oder kleiner als 0)
     unsigned short ux_short;
     unsigned int ux_int;
     unsigned long long ux_long_long;
@@ -213,24 +213,24 @@ int main (int argc, char** argv){
     '0'; // => 48 im ASCII-Zeichensatz
     'A'; // => 65 im ASCII-Zeichensatz
 
-    // sizeof(T) gibt die Grösse einer Variablen des Typen T in Bytes zurück.
-    // sizeof(obj) ergibt die Grösse des Ausdrucks (Variable, Literal usw.)
+    // sizeof(T) gibt die Größe einer Variablen des Typen T in Bytes zurück.
+    // sizeof(obj) ergibt die Größe des Ausdrucks (Variable, Literal usw.)
 
     printf("%zu\n", sizeof(int)); // => 4 (auf den Rechnern mit einem 4-Byte-Wort)
 
     // Wenn das Argument des `sizeof`-Operator ein Ausdruck ist, dann wird das
-    // Argument nicht ausgewertet (ausser Arrays mit variabler Länge)
+    // Argument nicht ausgewertet (außer Arrays mit variabler Länge)
     // Der Wert, der in diesem Fall zurückgegeben wird, ist eine Konstante zur 
     // Kompillierzeit.
 
     int a = 1;
     //size_t ist ein vorzeichenloser Integer Typ mit mindestens 2 Byte um die 
-    // Grösse eines Objekts zu repräsentieren.
+    // Größe eines Objekts zu repräsentieren.
     size_t size = sizeof(a++); // a++ wird nicht ausgewertet
     printf("sizeof(a++) = %zu, wobei a=%d ist\n", size, a);
     // Gibt "sizeof(a++) = 4, wobei a=1 ist" aus (mit einer 32-Bit-Architektur)
 
-    // Arrays müssen mit einer Grösse initialisiert werden.
+    // Arrays müssen mit einer Größe initialisiert werden.
     char my_char_array[20]; // Dieses Array beinhaltet 1 * 20 = 20 Bytes
     int my_int_array[20]; // Dieses Array beinhaltet 4 * 20 = 80 Bytes.
     // unter der Voraussetzung eines 4-Byte-Worts.
@@ -245,7 +245,7 @@ int main (int argc, char** argv){
     // Allerdings muss die Länge des Arrays dann zur Laufzeit ausgewertet werden:
     size_t my_array_size = sizeof(my_array) / sizeof(my_array[0]);
     // WARNUNG: Wenn dieser Ansatz gewählt wird, muss man sicherstellen, dass die
-    // Grösse des Arrays ermittelt werden *bevor* dieser einer Funktion als
+    // Größe des Arrays ermittelt werden *bevor* dieser einer Funktion als
     // Argument weitergegeben wird (siehe Diskussion weiter unten), weil Arrays
     // einer Funktion nur als Zeiger übergeben werden. => Das obere Statement
     // würde innerhalb einer Funktion ein falsches Resultat liefern.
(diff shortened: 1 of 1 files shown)


de-de/c-de.html.markdown
```diff
@@ -0,0 +1,880 @@
...
+contributors:
+    - ["caminsha", "https://github.com/caminsha"]
+lang: de-de
+---
+
+Ach, C. Immer noch **die** Sprache für modernes High-Performance Computing.
+
+C ist wahrscheinlich die niedrigste Programmiersprache, welche die meisten
+Programmierer je brauchen werden. Die Geschwindigkeit von C ist enorm, allerdings
+muss man sich stets der maneullen Speicherverwaltung bewusst sein.
```
mxschumacher: manuellen

de-de/c-de.html.markdown
```diff
@@ -0,0 +1,880 @@
...
+language: c
+filename: learnc-de.c
+contributors:
+    - ["caminsha", "https://github.com/caminsha"]
+lang: de-de
+---
+
+Ach, C. Immer noch **die** Sprache für modernes High-Performance Computing.
+
+C ist wahrscheinlich die niedrigste Programmiersprache, welche die meisten
```
mxschumacher: Sprache mit dem niedrigsten Abstraktionsniveau?

de-de/c-de.html.markdown
```diff
@@ -0,0 +1,880 @@
...
+muss man sich stets der maneullen Speicherverwaltung bewusst sein.
+
+
+> **Über Compiler Optionen**
+> 
+> Standardmässig sind `gcc` und `clang` ziemlich ruhig bezüglich Warnungen und 
+> Fehlern, obwohl dies sehr nützliche Informationen sein können. Es wird
+> empfohlen, strengere Compiler Optionen zu verwenden. Hier sind einige empfohlene
+> Standards:
+> `-Wall -Wextra -Werror -O2 -std=c99 -pedantic`
```
mxschumacher: should we stick to c99 or take a more recent standard?

de-de/c-de.html.markdown
```diff
@@ -0,0 +1,880 @@
...
+mehrzeilige Kommentare sehen so aus. Diese funktionieren auch in C89
+*/
+
+/* 
+mehrzeilige Kommentare können nicht verschachtelt werden /* Sei Vorsichtig! */ // Kommentar endet auf dieser Linie ...
+*/ // ... nicht bei dieser!
+
+// Konstanten: #define <keyword>
+// Konstanten werden laut der Konvention immer in GROSSBUCHSTABEN geschrieben
+#define TAGE_IM_JAHR 365
```
mxschumacher: even when programming in German, variable and function names themselves should still be in English

de-de/c-de.html.markdown
```diff
@@ -0,0 +1,880 @@
...
+// #include "mein_header.h"
+
+// Funktionssignaturen werden entweder vorher in einer .h-Datei deklariert oder
+// am Anfang der .c-Datei.
+void funktion_1();
+int funktion_2(void);
+
+// Es muss ein Funktionsprototyp deklariert werden vor der `main()` Funktion,
+// wenn die Funktion nach der `main()` Funktion gebraucht wird.
+int addiere_zwei_integer(int x1, int x2); // Funktionsprototyp
```
mxschumacher: again, I think function names should stay in English

de-de/c-de.html.markdown
```diff
@@ -0,0 +1,880 @@
...
+
+// Der Einstiegspunkt deines Programms ist eine Funktion mit dem Namen main und 
+// einem Integer als Rückgabewert.
+int main(void){
+    // dein Programm
+}
+
+// Die Kommandozeilenargumente, welche gebraucht werden, damit dein Programm 
+// läuft, werden als Argumente der `main`-Funktion mitgegeben.
+// argc steht für die Anzahl von Argumenten. - Der Programmname ist das erste
```
mxschumacher: for clarity, maybe you can mention that argc is "argument counter" and argv "argument vector"

de-de/c-de.html.markdown
```diff
@@ -0,0 +1,880 @@
...
+    // Typen
+    ////////////////////////////////////////////////
+
+    // Alle Variablen müssen am Anfang des jetzigen Blocks deklariert werden.
+    // Wir deklarieren die Variablen dynamisch im Code um die Lesbarkeit im
+    // Tutorial zu verbessern.
+    // C99-Konforme Compiler erlauben die Variablendeklaration an dem Punkt, an
+    // welchem die Variable verwendet wird.
+
+    // integer sind normalerweise 4 Bytes gross
```
mxschumacher: instead of "gross" it should be "groß" for all occurences

de-de/c-de.html.markdown
```diff
@@ -0,0 +1,880 @@
...
+    ////////////////////////////////////////////////
+    
+    // Kurzschreibweise für mehrere Deklarationen
+    int i1 = 1, i2 = 2;
+    flaot f1 = 1.0, f2 = 2.0;
+
+    int b,c;
+    b = c = 0;
+
+    // Arithmetik ist unkompliziert
```
mxschumacher: I believe basic arithmetic and logical operators should be closer to the beginning of the file, before arrays etc

The order in the Python article is quite healthy: https://learnxinyminutes.com/docs/python3/

de-de/c-de.html.markdown
```diff
@@ -0,0 +1,880 @@
...
+    i1 + i2; // => 3
```
mxschumacher: For clarity, we should use numbers themselves here and not variables. Doesn't add anything to the explanation to use

i1 and i2 instead of 1 + 2

de-de/c-de.html.markdown
```diff
@@ -0,0 +1,880 @@
...
+    // - Shift um einen Offset, welcher >= die Breite des linken Ausdrucks ist.
+    // int a = 1 << 32; // undefiniertes Verhalten, wenn int 32-Bit ist.
+    
+    ////////////////////////////////////////////////
+    // Kontrollstrukturen
+    ////////////////////////////////////////////////
+    if (0) {
+        printf("Ich werde nie ausgeführt.");
+    }
+    else if (0){
```
mxschumacher: let's be consistent about the style, in line 298 there is a space after if (0) and before the '{' in the next line, there isn't

I believe we should have a single space after the parens in the entire document

de-de/c-de.html.markdown
```diff
@@ -0,0 +1,880 @@
...
+    // wird, kann folgendes gemacht werden:
+    int i; 
+    for (i = 0; i <= 5; i++){
+        ; // Semikolon wird als Rumpf behandelt (Null-Anweisung)
+    }
+    // Alternativ kann auch folgendes geschrieben werden:
+    for (i = 0; i <= 5; i++);
+
+    // Verzweigungen mit mehreren Möglichkeiten: `switch()`
+    switch (a){
```
mxschumacher: see whitespace comment from above

de-de/c-de.html.markdown
```diff
@@ -0,0 +1,880 @@
...
+            printf("Schau mal ... 'a' ist entweder 3 oder 4.\n");
+            break;
+        default:
+            // wenn der Ausdruck `a` auf kein Label zutrifft.
+            fputs("Fehler!\n", stderr);
+            exit(-1);
+            break;
+    }
+
+    // Verwendung von "goto" in C
```
mxschumacher: given the famous debate around goto, I don't think this section should be included in the tutorial

de-de/c-de.html.markdown
caminsha: Didn't see this typo even though I read the whole article about 3 times. Thanks!

de-de/c-de.html.markdown
caminsha: Sounds better. I have changed it.

de-de/c-de.html.markdown
caminsha: Good question. I've heard that code written in C99 is more portable.

What do you think? Would C11 be better?

de-de/c-de.html.markdown
caminsha: Changed the variables and functions to English. Makes sense to me.

de-de/c-de.html.markdown
caminsha: Makes sense. Changed it.

de-de/c-de.html.markdown
caminsha: Ah, that's true. Didn't think about that because in Switzerland we don't know "ß" in words. 

Nevertheless I changed it, thanks!

de-de/c-de.html.markdown
caminsha: Removed this section. You're right it should not be part in a tutorial.

de-de/c-de.html.markdown
caminsha: True, makes it just more complicated.

de-de/c-de.html.markdown
caminsha: Sounds good to me. I didn't think about that.

In that case it should also be changed in the English tutorial, isn't it?

de-de/c-de.html.markdown
caminsha: I went through the content of the tutorial and searched for every word which contains `ss`and changed it to `ß`. 

To fix this I used the website duden.de.

de-de/c-de.html.markdown
mxschumacher: I cannot make that decision, probably project dependent. 

I believe it would be sensible to explain to provide a link to a resource that explains what those flags do. As things currently stand it is a bit confusing. Maybe this one? https://stackoverflow.com/questions/3375697/useful-gcc-flags-for-c

de-de/c-de.html.markdown
```diff
@@ -0,0 +1,866 @@
...
+["The C Programming Language"](https://de.wikipedia.org/wiki/The_C_Programming_Language) besorgt.
+Dieses Buch gilt als **das** Buch über die Programmiersprache C und wurde
+von Dennis Ritchie, dem Erfinder der Programmiersprache C, und Brian Kernighan
+geschrieben.
+Sei vorsichtig, da dieses Buch mittlerweile schon etwas älter ist und gewisse
+Unkorrektheiten (d.h. Ideen, welche nicht mehr als gut empfunden werden.) oder
+mittlerweile geänderte Praktiken enthält. [Hinweis: Das Buch wurde auf Englisch
+geschrieben, es gibt aber auch eine Übersetzung davon]
+
+Eine weitere gute Resource ist [Learn C The Hard Way](http://learncodethehardway.org/c/).
```
mxschumacher: Ressource

de-de/c-de.html.markdown
```diff
@@ -0,0 +1,866 @@
...
+[Englisch]
+
+Solltest du Fragen zu C haben, so lies die FAQ [compl.lang.c Frequently Asked Questions](http://c-faq.com).[Englisch]
+
+Außerdem ist es wichtig, eine saubere Einrückung zu verwenden. Des weiteren ist
+es wichtig, dass der Codestil möglichst konsistent ist. Es ist wichtiger, lesbaren
+Code zu schreiben als Code, welcher clever und schnell ist. Es lohnt sich ein
+Blick auf den [Codestil des Linuxkernel](https://www.kernel.org/doc/Documentation/process/coding-style.rst) zu werfen. [Englisch]
+
+Wenn die erwähnte Literatur dein Problem nicht löst, denke daran:
```
mxschumacher: I think we can delete these two lines

mxschumacher:
The PR is already much, much better. Thanks for all your work, @caminsha !

de-de/c-de.html.markdown
caminsha: I did not make a single space between parentheses but I made it more consistent that it looks like following snippet:

```c
for (i = 0; i < 10; i++){
    /* do something */
}
```

Is this also ok for you or would you prefer a space?

caminsha:
> The PR is already much, much better. Thanks for all your work, @caminsha !

Thanks for watching this PR! 

Are there still issues which should be removed?

de-de/c-de.html.markdown
mxschumacher: At the very bottom there is a link to the Linux Kernel C-Styleguide; if it is referenced here it is probably also a good idea to apply it in the document. They use spaces like this:

`for (int i = 0; i < 10; i++) {
     // do something
}`

caminsha: Ah that's true. I didn't read the Naming Conventions from the Kernel good enough.

I'll fix this in a next PR.

PR #3841, opened 2020-02-05
Fixed prototype function return value comment in javascript-ru

Badalik:


ru-ru/javascript-ru.html.markdown
```diff
@@ -420,7 +420,7 @@ myObj.__proto__ = myPrototype;
 myObj.meaningOfLife; // = 42
 
 // Для функций это тоже работает.
-myObj.myFunc(); // = "Привет, мир!"
+myObj.myFunc(); // = "привет, мир!"
```
mxschumacher: What's wrong with the current version?

Line 423 just returns what is defined in line 410, so a change would have to be made to both lines

Badalik: No, there will be a function return value with a toLowerCase from line 415

mxschumacher: ah yes, you are right.

PR #3842, opened 2020-02-06
[javascript/en] Add shorter way to define functions

rishinair11:
Add a shorter way to define functions with only one argument and a single statement in the function definition.

### DIFF
--- a/javascript.html.markdown
+++ b/javascript.html.markdown
@@ -622,7 +622,7 @@ const helloWorld = () => {
     console.log("Hello World")
 }
 
-// Also, I put the word "equivalent" in double quotes because a function defined
+// "Equivalent" refers to the fact that a function defined
 // using the arrow function cannnot be called before the definition.
 // The following is an example of invalid usage:
 
--- a/javascript.html.markdown
+++ b/javascript.html.markdown
@@ -611,15 +611,19 @@ pi = 4.13; // You cannot do this.
 
 const isOdd = number => number % 2 !== 0;
 
-// Parentheses are not required if only one argument is specified;
 // The "equivalent" of this function in the traditional syntax would look like this:
 
 function isOdd(number) {
     return number % 2 !== 0;
 };
 
-// I put the word "equivalent" in double quotes because a function defined
-// using the lambda syntax cannnot be called before the definition.
+// However if a function has no arguments, a slight change of adding the parenthesis in it's place is required:
+const helloWorld = () => {
+    console.log("Hello World")
+}
+
+// Also, I put the word "equivalent" in double quotes because a function defined
+// using the arrow function cannnot be called before the definition.
 // The following is an example of invalid usage:
 
 add(1, 8);
@@ -627,6 +631,7 @@ add(1, 8);
 const add = (firstNumber, secondNumber) => {
     return firstNumber + secondNumber;
 };
+
 ```
 
 ## Further Reading
--- a/javascript.html.markdown
+++ b/javascript.html.markdown
@@ -617,7 +617,7 @@ function isOdd(number) {
     return number % 2 !== 0;
 };
 
-// However if a function has no arguments, a slight change of adding the parenthesis in it's place is required:
+// However if a function has no arguments, a slight change of adding the parentheses in it's place is required:
 const helloWorld = () => {
     console.log("Hello World")
 }
--- a/javascript.html.markdown
+++ b/javascript.html.markdown
@@ -603,7 +603,8 @@ const pi = 3.14;
 
 pi = 4.13; // You cannot do this.
 
-// There is a new syntax for functions in ES6 known as "lambda syntax".
+// There is a new syntax for functions in ES6 known as "arrow function"
+// (lambda syntax in many others languages).
 // This allows functions to be defined in a lexical scope like with variables
 // defined by const and let. 
 // By default, a single statement defined will be the return statement, hence the `return` keyword is not required. 
--- a/javascript.html.markdown
+++ b/javascript.html.markdown
@@ -606,23 +606,15 @@ pi = 4.13; // You cannot do this.
 // There is a new syntax for functions in ES6 known as "lambda syntax".
 // This allows functions to be defined in a lexical scope like with variables
 // defined by const and let. 
-
-const isEven = (number) => {
-    return number % 2 === 0;
-};
-
-isEven(7); // false
-
-// An even shorter way to write a function similar to the above one is:
-// Parenthesis are not required if only one argument is specified;
-// By default, a single statement defined will be the return statement, hence `return` keyword is not required. 
+// By default, a single statement defined will be the return statement, hence the `return` keyword is not required. 
 
 const isOdd = number => number % 2 !== 0;
 
+// Parentheses are not required if only one argument is specified;
 // The "equivalent" of this function in the traditional syntax would look like this:
 
-function isEven(number) {
-    return number % 2 === 0;
+function isOdd(number) {
+    return number % 2 !== 0;
 };
 
 // I put the word "equivalent" in double quotes because a function defined
--- a/javascript.html.markdown
+++ b/javascript.html.markdown
@@ -613,6 +613,12 @@ const isEven = (number) => {
 
 isEven(7); // false
 
+// An even shorter way to write a function similar to the above one is:
+// Parenthesis are not required if only one argument is specified;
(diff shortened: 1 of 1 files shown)


javascript.html.markdown
```diff
@@ -613,6 +613,12 @@ const isEven = (number) => {
 
 isEven(7); // false
 
+// An even shorter way to write a function similar to the above one is:
```
mxschumacher: hey! feel free to delete the original isEven function (and adjust your comment accordingly). Your version is superior

javascript.html.markdown
```diff
@@ -613,6 +613,12 @@ const isEven = (number) => {
...
+// Parenthesis are not required if only one argument is specified;
+// By default, a single statement defined will be the return statement, hence `return` keyword is not required.
```
mxschumacher: add "the" in front of "`return`"

javascript.html.markdown
```diff
@@ -613,6 +613,12 @@ const isEven = (number) => {
 
 isEven(7); // false
 
+// An even shorter way to write a function similar to the above one is:
+// Parenthesis are not required if only one argument is specified;
```
mxschumacher: The plural of "parenthesis" is "Parentheses": https://data.grammarbook.com/blog/commas/all-about-parentheses/

javascript.html.markdown
```diff
@@ -606,17 +606,15 @@ pi = 4.13; // You cannot do this.
...
+// By default, a single statement defined will be the return statement, hence the `return` keyword is not required. 
 
-const isEven = (number) => {
-    return number % 2 === 0;
-};
-
-isEven(7); // false
+const isOdd = number => number % 2 !== 0;
 
+// Parentheses are not required if only one argument is specified;
```
kevinnls, line 613: however, they are required when there are no arguments
i feel we should demonstrate both
something like a simple
```
const sayHello = () => {
     console.log("Hello World") 
   }
```

rishinair11: Added in the following commit https://github.com/adambard/learnxinyminutes-docs/pull/3842/commits/9d11712b8bc5207763dd6546c4551fb76125f0a6

rishinair11:
Would someone kindly review and merge this PR into master?

rishinair11:
Would someone kindly review and merge this PR into master?

javascript.html.markdown
```diff
@@ -603,31 +603,35 @@ const pi = 3.14;
...
 };
 
-// I put the word "equivalent" in double quotes because a function defined
-// using the lambda syntax cannnot be called before the definition.
+// However if a function has no arguments, a slight change of adding the parentheses in it's place is required:
+const helloWorld = () => {
+    console.log("Hello World")
+}
+
+// Also, I put the word "equivalent" in double quotes because a function defined
```
kevinnls, line 625: i think the sudden personal tone is off-putting. i understand it's from a previous commit, but are there any ways you can think of rephrasing it?

kevinnls (approved):
looks all good to me. the other comment is only a suggestion. resolving not 100% required for merge. :+1:

javascript.html.markdown
```diff
@@ -603,31 +603,35 @@ const pi = 3.14;
 
 pi = 4.13; // You cannot do this.
 
-// There is a new syntax for functions in ES6 known as "lambda syntax".
+// There is a new syntax for functions in ES6 known as "arrow function"
+// (lambda syntax in many others languages).
 // This allows functions to be defined in a lexical scope like with variables
 // defined by const and let. 
+// By default, a single statement defined will be the return statement, hence the `return` keyword is not required.
```
nielsbom, line 610: Not exactly true:
`() => 3` will return 3, but 
`() => {3}` will not.

The rule here is: if (a) there's only one statement in the function body and (b) you drop the curly braces you can (1) no longer use the return statement and (2!) the resulting value of the statement is the resulting value of the function.

rishinair11:
Addressed the changes requested by @kevinnls and @nielsbom .
Good for merge?

javascript.html.markdown
```diff
@@ -603,31 +603,39 @@ const pi = 3.14;
 
 pi = 4.13; // You cannot do this.
 
-// There is a new syntax for functions in ES6 known as "lambda syntax".
+// There is a new syntax for functions in ES6 known as "arrow function"
+// (lambda syntax in many others languages).
 // This allows functions to be defined in a lexical scope like with variables
 // defined by const and let. 
+// By default, a single statement without '{ }' defined will be the return statement, hence the `return` keyword is not required.
```
nielsbom, line 610: This sentence is not grammatically correct and also kind of complicated.

Alternative:
> If the body of an arrow function does not have opening and closing braces (`{}`), it can only contain a single expression. The `return` keyword cannot be used in this case. The result of the given expression will be the return value of the function.

javascript.html.markdown
```diff
@@ -603,31 +603,39 @@ const pi = 3.14;
...
 // This allows functions to be defined in a lexical scope like with variables
 // defined by const and let. 
+// By default, a single statement without '{ }' defined will be the return statement, hence the `return` keyword is not required. 
 
-const isEven = (number) => {
-    return number % 2 === 0;
-};
+const isOdd = number => number % 2 !== 0;
+
+// But, if '{ }' are used, then a `return` is required.
```
nielsbom, line 614: Suggestion:
> If the body of an arrow function _is_ surrounded with braces (`{}`) then the function works as normal. If a return value is needed: use the `return` keyword.

nielsbom: "required" is too strong of a word here in my opinion, it's only required if you need a return value.

javascript.html.markdown
```diff
@@ -603,31 +603,39 @@ const pi = 3.14;
 
 pi = 4.13; // You cannot do this.
 
-// There is a new syntax for functions in ES6 known as "lambda syntax".
+// There is a new syntax for functions in ES6 known as "arrow function"
+// (lambda syntax in many others languages).
 // This allows functions to be defined in a lexical scope like with variables
 // defined by const and let.
```
nielsbom, line 609: I don't think this is true. I think you're conflating the concepts of "function expression" with "arrow function syntax".

We can have function **declarations**:
```
function add(a, b) {
  return a + b;
}
```

There are function **expressions**, in two flavors, both of which are expressions generating functions that can be assigned.
```
const add = function (a, b) {
  return a + b;
};
```

```
const add = (a, b) => a + b;
```

javascript.html.markdown
```diff
@@ -603,31 +603,39 @@ const pi = 3.14;
...
 
-isEven(7); // false
+const isEven = number => { return number % 2 == 0 }
 
 // The "equivalent" of this function in the traditional syntax would look like this:
 
-function isEven(number) {
-    return number % 2 === 0;
+function isOdd(number) {
+    return number % 2 !== 0;
```
nielsbom, line 621: These are not equivalent, they're completely different functions. Let's choose one (`isEven` is simpler) and use that multiple times.

javascript.html.markdown
```diff
@@ -603,31 +603,39 @@ const pi = 3.14;
...
 };
 
-// I put the word "equivalent" in double quotes because a function defined
-// using the lambda syntax cannnot be called before the definition.
+// However if a function has no arguments, a slight change of adding the parentheses in it's place is required:
+const helloWorld = () => {
+    console.log("Hello World")
+}
```
nielsbom, line 627: I'd introduce this more generally:

> If an arrow function has 0 arguments or more than 1 arguments it needs to have parentheses around its argument(s). If it has exactly 1 argument the parentheses can be omitted.

For example:

```
const giveMeFive = () => 5;
const addFive = number => number + 5;
const multiply = (number1, number2) => number1 * number2;
```

javascript.html.markdown
```diff
@@ -603,31 +603,39 @@ const pi = 3.14;
...
+
+// "Equivalent" refers to the fact that a function defined
+// using the arrow function cannnot be called before the definition.
```
nielsbom, line 630: To me this seems like "equivalent" is a meaningful word here, where it's actually not.

Instead I'd move this information to the part where we make the distinction between function declarations and function expressions and say something like:

> A function declaration _can_ be called before it's defined in a JavaScript file (this is a result of "hoisting").

```
add(1,8);

function add(number1, number2) {
  return number1 + number2;
}
```
> On the other hand a function expression _cannot_ be called before it's defined.

```
add(1,8); // error!

const add = function (number1, number2) {
  return number1 + number2;
};
```




```

```

PR #3862, opened 2020-02-19
[python/cs-cz] Fix bad description of floor division

tomasbedrich:

### DIFF
--- a/cs-cz/python.html.markdown
+++ b/cs-cz/python.html.markdown
@@ -42,7 +42,8 @@ Poznámka: Tento článek je zaměřen na Python 3. Zde se můžete [naučit sta
 # Až na dělení, které vrací desetinné číslo
 35 / 5  # => 7.0
 
-# Při celočíselném dělení je desetinná část oříznuta (pro kladná i záporná čísla)
+# Při celočíselném dělení je na výsledek aplikována funkce floor(),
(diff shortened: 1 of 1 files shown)


tomasbedrich:
Credit goes to Jirka Millek for noticing the error and notifying me.

cs-cz/python.html.markdown
```diff
@@ -42,7 +42,8 @@ Poznámka: Tento článek je zaměřen na Python 3. Zde se můžete [naučit sta
 # Až na dělení, které vrací desetinné číslo
 35 / 5  # => 7.0
 
-# Při celočíselném dělení je desetinná část oříznuta (pro kladná i záporná čísla)
+# Při celočíselném dělení je na výsledek aplikována funkce floor(),
+# což znamená zaokrouhlení směrem k mínus nekonečnu (pro kladná i záporná čísla).
```
mxschumacher: Using google Translate I get:

In integer division, the floor () function is applied to the result, which means rounding to minus infinity (for both positive and negative numbers).

The python documentation says (https://docs.python.org/3/library/math.html):

Return the floor of x, the largest integer less than or equal to x.

tomasbedrich: The translation is wrong, the exact translation would be: "... which means rounding towards minus infinity ...".

Although the documentation defines it precisely, I believe that this alternative formulation tells the same message and it seems to me as more human-friendly.

Issue #3946, opened 2020-06-01
Duplicate 'vim' entries on the main page


anakimluke:
![image](https://user-images.githubusercontent.com/22756345/83461177-65bc7a80-a43e-11ea-9d60-0fd2532fd1d1.png)

The first entry links to https://learnxinyminutes.com/docs/el-gr/vim-gr/

And the second one links to https://learnxinyminutes.com/docs/vim/

Notice how on the first 'vim' entry row, there is no `gr-el` option to be clicked on.

Upon entering the first vim link, the greek(?) vim doc page, parts of the page are in english, and some are in greek.

![image](https://user-images.githubusercontent.com/22756345/83461414-10cd3400-a43f-11ea-8349-a814359df940.png)

whereas it should all be in greek.

PR #3947, opened 2020-06-01
Fixed duplicate vim entry and page language mismatch

anakimluke:
This refers to #3946 
I've not marked the title with `[language/lang-code]` so as to make clear it's a meta issue.

I built the project and can confirm this fixes the issue:

There's now only one vim entry:
image

and the `el-gr` button now exists.

Both `vim`  and `el-gr` now link to the right places.

And the language of the whole page now matches the selected(greek) langugage:
image

### DIFF
--- a/el-gr/vim-gr.html.markdown
+++ b/el-gr/vim-gr.html.markdown
@@ -4,7 +4,7 @@ tool: vim
 contributors:
     - ["RadhikaG", "https://github.com/RadhikaG"]
 filename: LearnVim.txt
-land: el-gr
+lang: el-gr
 ---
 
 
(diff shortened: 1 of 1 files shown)


el-gr/vim-gr.html.markdown
```diff
@@ -4,6 +4,7 @@ tool: vim
 contributors:
     - ["RadhikaG", "https://github.com/RadhikaG"]
 filename: LearnVim.txt
+land: el-gr
```
imba-tjd, line 7:
```suggestion
lang: el-gr
```

anakimluke: i can't believe this

anakimluke:
@imba-tjd In my defense, I *did* build the project with what must have been the typo-less change. I just did not commit that change, and instead edited the file again in the browser. 😅

PR #3988, opened 2020-08-03
[Powershell/en] Refactoring Powershell Wiki

AndrewDavis1191:
Changing to reflect current capabilities and draw similarities to other languages in terms of capabilities

### DIFF
--- a/powershell.html.markdown
+++ b/powershell.html.markdown
@@ -18,7 +18,7 @@ rather than plain text. After years of evolving, it resembles Python a bit.
 
 [Read more here.](https://docs.microsoft.com/powershell/scripting/overview)
 
-<H3>Powershell as a Language:</H3>
+Powershell as a Language:
 
 ```powershell
 
@@ -645,7 +645,7 @@ function Format-Range ($start, $end) {
 
 Format-Range 2 6 # => 'a','b','g','f','e','d','c','h','i','j','k','l','m','n'
 ```
-<H3>Powershell as a Tool:</H3>
+Powershell as a Tool:
 
 Getting Help:
 
--- a/powershell.html.markdown
+++ b/powershell.html.markdown
@@ -51,7 +51,8 @@ Powershell as a Language:
 [int]-5 / [int]3      # => -1.66666666666667
 5.0 / 3.0   # => 1.66666666666667
 -5.0 / 3.0  # => -1.66666666666667
-[int]$result = 5 / 3 # => 2
+[int]$result = 5 / 3 
+$result # => 2
 
 # Modulo operation
 7 % 3  # => 1
@@ -716,8 +717,8 @@ C:\Program Files\dotnet\dotnet.exe    # returns string rather than execute
 # One usage of '&' would be to invoke a scriptblock inside of your script.
 # Notice the variables are scoped
 $i = 2
-$scriptblock = { $i=5; Write-Output $i }
-& $scriptblock # => 5
+$scriptBlock = { $i=5; Write-Output $i }
+& $scriptBlock # => 5
 $i # => 2
 
 invoke-expression ' $i=5; Write-Output $i ' # => 5
--- a/powershell.html.markdown
+++ b/powershell.html.markdown
@@ -183,11 +183,12 @@ $someVariable  # => 5
 # Ternary Operators exist in Powershell 7 and up
 0 ? 'yes' : 'no'  # => no
 
-# The default array object in Powershell is an immutable array
+# The default array object in Powershell is an fixed length array
 $defaultArray = "thing","thing2","thing3"
-# you are unable to add or remove objects
+# you can add objects with '+=', but cannot remove objects
 $defaultArray.Add("thing4") # => Exception "Collection was of a fixed size."
-# To have a mutable array, you will need to use the .NET ArrayList class
+# To have a more workable array, you'll want the .NET [ArrayList] class
+# It is also worth noting that ArrayLists are significantly faster
 
 # ArrayLists store sequences
 [System.Collections.ArrayList]$array = @()
--- a/powershell.html.markdown
+++ b/powershell.html.markdown
@@ -89,12 +89,12 @@ $False - 5   # => -5
 
 # Using boolean logical operators on ints casts them to booleans for evaluation
 # but their non-cast value is returned
-# Don't mix up with bool(ints) and bitwise and/or (&,|)
+# Don't mix up with bool(ints) and bitwise -band/-bor
 [bool](0)     # => False
 [bool](4)     # => True
 [bool](-6)    # => True
-0 -and 2     # => 0
--5 -or 0     # => -5
+0 -band 2     # => 0
+-5 -bor 0     # => -5
 
 # Equality is -eq (equals)
 1 -eq 1  # => True
--- a/powershell.html.markdown
+++ b/powershell.html.markdown
@@ -19,27 +19,8 @@ rather than plain text. After years of evolving, it resembles Python a bit.
 [Read more here.](https://docs.microsoft.com/powershell/scripting/overview)
 
 Powershell as a Language:
+
 ```powershell
-
-# Single line comments start with a number symbol.
-
-<#
-  Multi-line comments
-  like so
-#>
-
-####################################################
-## 1. Primitive Datatypes and Operators
-####################################################
-
-# Numbers
-3 # => 3
-
-# Math
-1 + 1   # => 2
-8 - 1   # => 7
-10 * 2  # => 20
-35 / 5  # => 7.0
 
 # Single line comments start with a number symbol.
 
--- a/powershell.html.markdown
+++ b/powershell.html.markdown
@@ -43,9 +43,10 @@ Powershell as a Language:
 10 * 2  # => 20
 35 / 5  # => 7.0
 
-# Powershell uses banker's rounding
-# Meaning [int]1.5 would round to 2 but so would [int]2.5
-# division always returns a float. You must cast result to [int] to round
+# Powershell uses banker's rounding,
+# meaning [int]1.5 would round to 2 but so would [int]2.5
+# Division always returns a float. 
+# You must cast result to [int] to round.
 [int]5 / [int]3       # => 1.66666666666667
 [int]-5 / [int]3      # => -1.66666666666667
 5.0 / 3.0   # => 1.66666666666667
@@ -55,10 +56,10 @@ Powershell as a Language:
 # Modulo operation
 7 % 3  # => 1
 
-# Exponentiation requires longform or the built-in [Math] class
+# Exponentiation requires longform or the built-in [Math] class.
 [Math]::Pow(2,3)  # => 8
 
-# Enforce order of operations with parentheses
+# Enforce order of operations with parentheses.
 1 + 3 * 2  # => 7
 (1 + 3) * 2  # => 8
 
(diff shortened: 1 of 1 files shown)


AndrewDavis1191:
@divayprakash are you able to review this? :innocent:

divayprakash:
@AndrewDavis1191 Sorry I don't knwo any Powershell so won't be able to review such a large change.

imba-tjd:
Is this expected?

图片

AndrewDavis1191:
> ~Is this expected?~ I guess not. There are two `1. Primitive Datatypes and Operators`

Uh oh, good catch. I totally missed that. Removed in e2b7568

powershell.html.markdown
```diff
@@ -13,116 +14,348 @@ Nearly all examples below can be a part of a shell script or executed directly
...
+-5 -ne $False # => True
+
+# Using boolean logical operators on ints casts them to booleans for evaluation
+# but their non-cast value is returned
+# Don't mix up with bool(ints) and bitwise and/or (&,|)
+[bool](0)     # => False
+[bool](4)     # => True
+[bool](-6)    # => True
+0 -and 2     # => 0
+-5 -or 0     # => -5
```
imba-tjd, line 97: My output is different

powershell.html.markdown
```diff
@@ -13,116 +14,348 @@ Nearly all examples below can be a part of a shell script or executed directly
...
+"Hello world!"[0]  # => 'H'
+
+# You can find the length of a string
+("This is a string").Length  # => 16
+
+# You can also format using f-strings or formatted string literals
+$name = "Steve"
+$age = 22
+"He said his name is $name." # => "He said his name is Steve"
+“{0} said he is {1} years old.” -f $name, $age # => "Steve said he is 22 years old"
```
imba-tjd, line 146:
```suggestion
"{0} said he is {1} years old." -f $name, $age # => "Steve said he is 22 years old"
```

powershell.html.markdown
AndrewDavis1191: Needed to change to -band/-bor for posh. Changed in 998249d

powershell.html.markdown
```diff
@@ -13,116 +14,349 @@ Nearly all examples below can be a part of a shell script or executed directly
...
+# $null, 0, and empty strings and arrays all evaluate to False.
+# All other values are True
+function test ($value) {
+  if ($value) {Write-Output 'True'}
+    else {Write-Output 'False'}
+}
+test ($null) # => False
+test (0)   # => False
+test ("")  # => False
+test []  # => True
```
imba-tjd, line 161: I think this is confusing.

imba-tjd:
```
  { 'x' -contains 'x'}    { "FALSE! -contains is for lists!"; break }
```

`'x' -contains 'x'` returns True.

powershell.html.markdown
AndrewDavis1191: I've added some comments and details to the objects.

AndrewDavis1191:
> ```
>   { 'x' -contains 'x'}    { "FALSE! -contains is for lists!"; break }
> ```
> 
> `'x' -contains 'x'` returns True.

This was from the original creator. It may have been true at that time, but it seems it is not true in the current version of Powershell.

imba-tjd:
```
function test ($value) { $value.gettype() }
test []
```

The result is string

powershell.html.markdown
```diff
@@ -13,116 +14,373 @@ Nearly all examples below can be a part of a shell script or executed directly
...
+"{0} said he is {1} years old." -f $name, $age # => "Steve said he is 22 years old"
+"$name's name is $($name.Length) characters long." # => "Steve's name is 5 characters long."
+
+# $null is not an object
+$null  # => None
+
+# $null, 0, and empty strings and arrays all evaluate to False.
+# All other values are True
+# $null, 0, and empty strings and arrays all evaluate to False.
+# All other values are True
```
imba-tjd, line 155: You know

powershell.html.markdown
```diff
@@ -13,116 +14,373 @@ Nearly all examples below can be a part of a shell script or executed directly
...
+You cannot call a method on a null-valued expression.
+At line:4 char:1
++ $test3.getType()
++ ~~~~~~~~~~~~~~~~
+    + CategoryInfo          : InvalidOperation: (:) [], RuntimeException
+    + FullyQualifiedErrorId : InvokeMethodOnNull
+ 
+True     True     ScriptBlock                              System.Object                                          
+True     True     Object[]                                 System.Array  
+#>
```
imba-tjd, line 187: same here

AndrewDavis1191:
> ```
> function test ($value) { $value.gettype() }
> test []
> ```
> 
> The result is string
That's interesting, a null-valued expression seems to be converted to a string when passed through a function.
> 
> And I don't know whether you keep L172-L187 intentionally.
I did, but it seems unnecessary I suppose.
> 
> BTW I will need some days to review this because my mom only allows me to use pc for a limited time and I'm also a beginner of PS.
No worries, take your time.

powershell.html.markdown
```diff
@@ -133,89 +373,54 @@ switch($val) {
...
 
-### Functions
-# The [string] attribute is optional.
-# Function names should follow Verb-Noun convention
-function Get-Foo([string]$name) {
-	echo "Hey $name, have a function"
+# Calling functions with parameters
+function Add-ParamNumbers {
+ param( [int]$FirstNumber, [int]$SecondNumber )
+ $FirstNumber + $SecondNumber
```
imba-tjd, line 420: How about using `Add-ParamNumbers([int]$FirstNumber, [int]$SecondNumber)`, which would make it more similar to other language.

AndrewDavis1191: There are later examples that illustrate this. I think in production you'd just go with whatever your team does.

imba-tjd:
I have finished reading. My thoughts:

1. covering the usage of `&"xxx.exe"`. In any other shell language, `"xxx.exe"` is enough to call an external program, but for PS it only returns the original string.
2. When it refers to .net classes, the `System` can be omitted, e.g. `[System.Math]` -> `[Math]`
3. In `Show-Command Get-EventLog`, the `Get-EventLog` isn't exist in pwsh7.
4. PS doesn't support raw(bytes) piping: `curl.exe https://www.microsoft.com/favicon.ico -O` works, but `curl.exe https://www.microsoft.com/favicon.ico > favicon.ico` will break the image. It's worth mentioning this pitfall.

AndrewDavis1191:
> 1. covering the usage of `&"xxx.exe"`. In any other shell language, `"xxx.exe"` is enough to call an external program, but for PS it only returns the original string.

I think in most cases, just typing the name works fine, but I see what you are referring to. If this will be included, then I think we should include dot sourcing as well for info on scoping the external commands run in this fashion.

> 2. When it refers to .net classes, the `System` can be omitted, e.g. `[System.Math]` -> `[Math]`

True, but I think in a tutorial we should just mention that in a comment. In the cases like [System.Collections.ArrayList] I think it is informative to see the drilldown in that way.

> 3. In `Show-Command Get-EventLog`, the `Get-EventLog` doesn't exist in pwsh7.

Ah yes, looks like it's deprecated in favor of "Get-WinEvent" [Issue #9852 on Powershell repo](https://github.com/PowerShell/PowerShell/issues/9852)

> 4. PS doesn't support raw(bytes) piping: `curl.exe https://www.microsoft.com/favicon.ico -O` works, but `curl.exe https://www.microsoft.com/favicon.ico > favicon.ico` will break the image. It's worth mentioning this pitfall.

Good idea, I don't think I've ever run into this issue personally; didn't even realize it was an issue.
I see there is and has been an open issue on this for quite some time. [Issue #1908 on Powershell repo](https://github.com/PowerShell/PowerShell/issues/1908)

> 5. I prefer keeping `1..10 | % { "Loop number $_" }` and some of the `| ?` because I think it's important to mention the pipeline idea.

I agree somewhat. Typically the pipeline is slower than alternative routes, but the convenience is invaluable. I would not use the % or ? or $_ aliases and instead recommend using the full command. It is more readable in my opinion.

AndrewDavis1191:
Ah I see. The CI is failing currently due to a lint on [vim/de](https://github.com/adambard/learnxinyminutes-docs/pull/3999)

powershell.html.markdown
```diff
@@ -554,6 +556,16 @@ True     False    Guitar                                   Instrument
 #>
 Get-Process | Foreach-Object ProcessName | Group-Object
 
+# Useful pipeline examples are iteration and filtering
+1..10 | ForEach-Object { "Loop number $PSITEM" }
+1..10 | where {$PSITEM -gt 5} | Format-Table
+
+# A noteable pitfall of the pipeline is it's performance when
+# compared with other options
+# additionally, raw bytes are not passed through the piipeline
```
imba-tjd, line 565: typo: piipeline

AndrewDavis1191: Gotta love the Drop mech keyboard quality.

imba-tjd:
The point 1, my idea is that when the path contains spaces, in cmd you just need to add quotation marks:

```
> C:\Program Files\dotnet\dotnet.exe
'C:\Program' is not recognized as an internal or external command,
operable program or batch file.

> "C:\Program Files\dotnet\dotnet.exe"
...
(works)
```

But this experience doesn't work in PS:

```
> C:\Program Files\dotnet\dotnet.exe
C:\Program: The term 'C:\Program' is not recognized as a name of a cmdlet, function, script file, or executable program.
Check the spelling of the name, or if a path was included, verify that the path is correct and try again

> "C:\Program Files\dotnet\dotnet.exe"
C:\Program Files\dotnet\dotnet.exe    # returns the string rather than execute it
```

Besides the arguments shouldn't be included in quotation marks:

```
&"C:\Program Files\dotnet\dotnet.exe --help"   # fail
&"C:\Program Files\dotnet\dotnet.exe" --help   # success
```

I think this is a common case when writing scripts.

Besides, I now find another thing worthy to mention. PS is using backticks to escape characters, which is different from most of the other languages. The benefit is that you needn't use `\\` in strings which is the directory seperator in Windows.

AndrewDavis1191:
> The point 1, my idea is that when the path contains spaces, in cmd you just need to add quotation marks:

Ah, I see what you mean. I can add info on calling external exe's as well with the spaces

> ```
> &"C:\Program Files\dotnet\dotnet.exe --help"   # fail
> &"C:\Program Files\dotnet\dotnet.exe" --help   # success
> ```

For this purpose we can also mention dot-sourcing as an option
```
."C:\Program Files\dotnet\dotnet.exe" --help
```

> Besides, I now find another thing worthy to mention. PS is using backticks to escape characters, which is different from most of the other languages. The benefit is that you needn't (and couldn't) use `\\` in strings which is the directory seperator in Windows.

I intentionally left out any mention of backticks, since I just think they are ugly, and there are many workarounds for their uses that are much nicer. The escape chars have no alternative afaik though, so I'll include info on those.

powershell.html.markdown
```diff
@@ -149,6 +149,18 @@ $age = 22
...
+# Escape Characters in Powershell
+# Many languages use the '\', but Windows uses this character for 
+# file paths. Powershell thus uses '`' to escape characters
+# Take caution when working with files, as '`' is a
+# valid character in NTFS filenames.
+"Showing`nEscape Chars" # => new line between Showing and Escape
+"Making`tTables`tWith`tTabs" # => Format things with tabs
+
+# Negate pound sign to prevent comment
+# Note that the function of '#' is removed, but '#" is still present
```
imba-tjd, line 161: Could you elaborate this line?

powershell.html.markdown
```diff
@@ -13,116 +14,368 @@ Nearly all examples below can be a part of a shell script or executed directly
...
+2 -lt 3 -and 3 -lt 2  # => False
+
+# (-is vs. -eq) -is checks if two objects are the same type
+# -eq checks if the objects have the same values.
+# Note: we called '[Math]' from .NET previously without the preceeding
+# namespaces. We can do the same with [Collections.ArrayList] if preferred
+[System.Collections.ArrayList]$a = @()  # Point a at a new list
+$a = (1,2,3,4)
+$b = $a                                 # => Point b at what a is pointing to
+$b -is $a.getType()                     # => True, a and b equal same type
```
imba-tjd, line 125:
```
GetType
```

imba-tjd (approved):
🆗 That's all my reviews.

AndrewDavis1191:
@imba-tjd I really appreciate you taking the time to review this! Thank you! :Grin:

PR #3997, opened 2020-08-16
[js/fr] fix for...of loop

### LINKED ISSUE #3894: [javascript/fr] Example with for..of on a object literal => TypeError: ... not iterable !
Hey !
There is a little problem on the French version of the JavaScript page (this part is okay on the English version)
The problematic part is : 
```
// *ES6:* La boucle for...of permet d'itérer sur les propriétés d'un objet
var description = "";
var person = {fname:"Paul", lname:"Ken", age:18};
for (var x of person){
    description += x + " ";
}
description; // = "Paul Ken 18 "
```
When you try this, you will have an error :  **Uncaught TypeError: person is not iterable**
And it is normal because a object literal is not iterable !

To fix, it could be possible to just translate the english part like this (and sorry but i'm not sure how to fix it myself) : 
```
// La boucle for...of permet de parcourir un objet itérable
// (ce qui inclut les objets Array, Map, Set, String, ... Mais pas un objet littéral !)
var myPets = "";
var pets = ["cat", "dog", "hamster", "hedgehog"];
for (var pet of pets){
    myPets += pet + " ";
} // myPets = 'cat dog hamster hedgehog '
```

In any case, thanks for the site and all of this ! =)

/Carl

kevinnls:
Co-authored-by: Carl20cent <carl.vincent38@gmail.com>
resolves and closes #3894
***
- [x] I solemnly swear that this is all ~~original~~ improved content of which @Carl20cent is the original author (and i made a few adjustments)

### DIFF
--- a/fr-fr/javascript-fr.html.markdown
+++ b/fr-fr/javascript-fr.html.markdown
@@ -332,7 +332,7 @@ description; // = "Paul Ken 18 "
 // (ce qui inclut les objets Array, Map, Set, String, ... Mais pas un objet littéral !)
 let myPets = "";
 const pets = ["cat", "dog", "hamster", "hedgehog"];
-for (let pet of pets){ //`(const pet of pets)` is also possible
+for (let pet of pets){ //`(const pet of pets)` est également possible
 
     myPets += pet + " ";
 }
--- a/fr-fr/javascript-fr.html.markdown
+++ b/fr-fr/javascript-fr.html.markdown
@@ -1,12 +1,12 @@
 ---
 language: javascript
 contributors:
-    - ['Leigh Brenecki', 'https://leigh.net.au']
-    - ['Ariel Krakowski', 'http://www.learneroo.com']
+    - ["Leigh Brenecki", "https://leigh.net.au"]
+    - ["Ariel Krakowski", "http://www.learneroo.com"]
 filename: javascript-fr.js
 translators:
-    - ['@nbrugneaux', 'https://nicolasbrugneaux.me']
-    - ['Michel Antoine', 'https://github.com/antoin-m']
+    - ["@nbrugneaux", "https://nicolasbrugneaux.me"]
+    - ["Michel Antoine", "https://github.com/antoin-m"]
 lang: fr-fr
 ---
 
@@ -328,13 +328,15 @@ for (var x in person){
 }
 description; // = "Paul Ken 18 "
 
-// *ES6:* La boucle for...of permet d'itérer sur les propriétés d'un objet
-var description = "";
-var person = {fname:"Paul", lname:"Ken", age:18};
-for (var x of person){
-    description += x + " ";
+// *ES6:* La boucle for...of permet de parcourir un objet itérable
+// (ce qui inclut les objets Array, Map, Set, String, ... Mais pas un objet littéral !)
+let myPets = "";
+const pets = ["cat", "dog", "hamster", "hedgehog"];
+for (let pet of pets){ //`(const pet of pets)` is also possible
+
+    myPets += pet + " ";
 }
-description; // = "Paul Ken 18 "
+myPets; // = 'cat dog hamster hedgehog '
 
 // && est le "et" logique, || est le "ou" logique
 if (house.size === 'big' && house.colour === 'blue'){


kevinnls:
@Carl20cent @mxschumacher take a look~!

fr-fr/javascript-fr.html.markdown
```diff
@@ -328,13 +328,14 @@ for (var x in person){
...
-// *ES6:* La boucle for...of permet d'itérer sur les propriétés d'un objet
-var description = "";
-var person = {fname:"Paul", lname:"Ken", age:18};
-for (var x of person){
-    description += x + " ";
+// *ES6:* La boucle for...of permet de parcourir un objet itérable
+// (ce qui inclut les objets Array, Map, Set, String, ... Mais pas un objet littéral !)
+var myPets = "";
+var pets = ["cat", "dog", "hamster", "hedgehog"];
+for (var pet of pets){
```
mxschumacher, line 335: we can use const here

fr-fr/javascript-fr.html.markdown
```diff
@@ -328,13 +328,14 @@ for (var x in person){
...
 description; // = "Paul Ken 18 "
 
-// *ES6:* La boucle for...of permet d'itérer sur les propriétés d'un objet
-var description = "";
-var person = {fname:"Paul", lname:"Ken", age:18};
-for (var x of person){
-    description += x + " ";
+// *ES6:* La boucle for...of permet de parcourir un objet itérable
+// (ce qui inclut les objets Array, Map, Set, String, ... Mais pas un objet littéral !)
+var myPets = "";
```
mxschumacher, line 333: let's use "let", var should be avoided in any new code

fr-fr/javascript-fr.html.markdown
```diff
@@ -1,12 +1,12 @@
 ---
 language: javascript
 contributors:
-    - ['Leigh Brenecki', 'https://leigh.net.au']
```
mxschumacher, line 4: changing the quotes seems to be an unrelated change that doesn't improve the code

kevinnls: it conforms with the specs on [CONTRIBUTING.md](https://github.com/adambard/learnxinyminutes-docs/blob/master/CONTRIBUTING.markdown#header-configuration)

(truth: my formatter did it)

fr-fr/javascript-fr.html.markdown
kevinnls: updated at 4c9e1c8

fr-fr/javascript-fr.html.markdown
kevinnls: updated at 4c9e1c8

kevinnls:
@mxschumacher at 4c9e1c8 i've changed the variable declarations in the section that we're rectifying rn. the whole document would require more time and i don't think i could take that up rn.

fr-fr/javascript-fr.html.markdown
mxschumacher: okay, then we can use double quotes

fr-fr/javascript-fr.html.markdown
```diff
@@ -328,13 +328,14 @@ for (var x in person){
...
-// *ES6:* La boucle for...of permet d'itérer sur les propriétés d'un objet
-var description = "";
-var person = {fname:"Paul", lname:"Ken", age:18};
-for (var x of person){
-    description += x + " ";
+// *ES6:* La boucle for...of permet de parcourir un objet itérable
+// (ce qui inclut les objets Array, Map, Set, String, ... Mais pas un objet littéral !)
+let myPets = "";
+const pets = ["cat", "dog", "hamster", "hedgehog"];
+for (let pet of pets){
```
mxschumacher, line 335: as I have said before, in the for .. of loop you can use "const", see here: https://stackoverflow.com/questions/31987465/ecmascript-2015-const-in-for-loops

kevinnls: ah thanks! i didn't know that.

but even so, i see no particular logical reason to choose either `const` or `let` – they both function in the same way. this leaves it as a choice based on style and conventions<sup>1</sup> which should be left to the dev. so may i suggest something like this?  


```suggestion
for (let pet of pets){ //`(const pet of pets)` is also possible

```
\
or is there any reason for const to enforced?
***
1.the convention as far as i have seen appears to be `let` (it's one letter fewer). but i may be wrong

mxschumacher: As a general rule, I suggest to be as restrictive as possible - if you can use const, use const. 

Here's another elaboration: https://exploringjs.com/es6/ch_for-of.html#sec_iteration-variable-for-of

Saving one letter isn't important here - what the JS engine does when reading the code matters.

I think it should just be.

`for (const pet of pets)`

kevinnls:
i may be wrong and @adambard could correct me for it. sorry, but i don't think LXIYM was started with the intent of being an authoritative text. but i don't see why we need to decide for the dev who reads that `const` is the best approach rather than inform them that both approaches are possible and leave it to them to figure out which works best for them. i stick by my decision to include both `let` and `const` in the document

adambard:
I've seen people argue both for and against pervasive use of `const`, so including both seems fair enough to me (my personal preference would be to represent const as the default, but I don't think it's worth making a big deal of either way).

kevinnls:
sounds good. thank you @adambard and @mxschumacher

PR #4033, opened 2020-10-06
[python/en] adding an example for closures in python where we use the `nonlocal` …

Mohammed-Ashour:
…keyword

Adding an example for the case of nested functions with closure scopes, when we use the python 3s keyword 'nonlocal' to point to the variables in the outer functions

### DIFF
--- a/python.html.markdown
+++ b/python.html.markdown
@@ -635,6 +635,21 @@ def create_adder(x):
 add_10 = create_adder(10)
 add_10(3)   # => 13
 
+# Closures in nested functions, the nonlocal keyword to work with variables in nested scope which shouldn't be in the inner functions.
+def create_avg():
+    total = 0
+    count = 0
+    def avg(n):
+        nonlocal total, count
+        total += n
+        count += 1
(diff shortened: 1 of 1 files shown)


python.html.markdown
```diff
@@ -635,6 +635,21 @@ def create_adder(x):
 add_10 = create_adder(10)
 add_10(3)   # => 13
 
+# Closures in nested functions, the nonlocal keyword to work with variables in nested scope which shouldn't be in the inner functions.
```
vendethiel, line 638: Prefer a statement, like:
```python
# Closures in nested functions:
# We can use the nonlocal keyword to work with variables in nested scope which shouldn't be declared in the inner functions.
```

Mohammed-Ashour: Addressed in https://github.com/adambard/learnxinyminutes-docs/pull/4033/commits/28b4f73f1bc3055df8b0ca00dc49e93d0ced2f20

vendethiel:
Thanks

PR #4035, opened 2020-10-11
[Dart/en] Fixed error in example2

AntonTrekov:
The current example is not working.
The code is slightly different from examples from other languages. I wanted to continue the initial thought and show that you can specify arguments and returned type.
An example without using a type indication is used in the following example.

- [ ] I solemnly swear that this is all original content of which I am the original author
- [x] Pull request title is prepended with [language/lang-code]
- [x] Content changes are aimed at intermediate to experienced programmers (this is a poor format for explaining fundamental programming concepts)
- [x] If you've changed any part of the YAML Frontmatter, make sure it is formatted according to CONTRIBUTING.md

### DIFF
--- a/dart.html.markdown
+++ b/dart.html.markdown
@@ -77,10 +77,10 @@ example1() {
   nested1();
 }
 
-/// Anonymous functions don't include a name but can take number of arguments
+/// Anonymous functions don't include a name
+/// but can take number of arguments and return type
 example2() {
-  //// Explicit return type.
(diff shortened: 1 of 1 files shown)


AntonTrekov:
Not sure if such changes can be considered original content

AntonTrekov:
I decided that it would be better to do everything the same as in the examples for other languages, so that there were no conflicts

geoffliu (approved):

PR #4060, opened 2020-11-21
[rust/el-gr] Small fixes to Rust el-gr

amilios:
Fixed some orthographic mistakes (βρόγχοι instead of the correct βρόχοι, ανοίκει instead of the correct ανήκει). The original poster also used the phrase "μη-αμετάβλητο" (non-immutable), which I just replaced with μεταβλητό (mutable). 

Reworded some of the explanations for mutable vs. immutable variables. These concepts are somewhat hard to translate into Greek; a naive translation is μεταβλητές μεταβλητές and αμετάβλητες μεταβλητές, however that sounds very awkward, as the word for "variable" itself in Greek is the noun-ified form of the adjective "mutable", so it is like saying "mutable mutable" and "immutable mutable" -- it doesn't really make sense, and the duplication sounds weird. I get around this by specifying αμετάβλητη δέσμευση - immutable binding, but referring to immutable variables still sounds weird (even in English it kind of sounds weird to me -- a variable that doesn't vary?). Unfortunately there is not enough Greek documentation for Rust (not at all), for me to figure out what the accepted translation is. 


vendethiel:
Thanks!

PR #4075, opened 2020-12-23
[c/en] Correct information about `sizeof`

mohamedhaddi:
Correct information about "`sizeof` only evaluating an expression if it's a VLA."
When trying this, it seems that `sizeof` would evaluate even arrays of constant known sizes, not just VLAs:

```c
int size = 42;
int array1[size];
int array2[42];

printf("%zu\n", sizeof(array1)); // => 168
printf("%zu\n", sizeof(array2)); // => 168
```

### DIFF
--- a/c.html.markdown
+++ b/c.html.markdown
@@ -131,7 +131,7 @@ int main (int argc, char** argv)
   printf("%zu\n", sizeof(int)); // => 4 (on most machines with 4-byte words)
 
   // If the argument of the `sizeof` operator is an expression, then its argument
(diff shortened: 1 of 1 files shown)


imba-tjd:
I would vote down, thouth I don't have much experience in C.

https://en.cppreference.com/w/c/language/sizeof
> Except if the type of expression is a VLA, (since C99)expression is not evaluated

mohamedhaddi:
Sorry, I misunderstood what was meant by an expression... This answer clarifies it very well:
https://stackoverflow.com/a/48661592/4273342

PR #4083, opened 2021-01-09
[elixir/en] Fix a few 404 references on the Elixir page

oskarrough:
There are two 404 links on the English Elixir page. This fixes them.


oskarrough:
The links were changed to search queries live. Seems OK.

PR #4084, opened 2021-01-11
[zfs/ru] Add russian translation for ZFS doc

A1EF:

### DIFF
--- a/ru-ru/zfs-ru.html.markdown
+++ b/ru-ru/zfs-ru.html.markdown
@@ -1,0 +1,406 @@
+---
+category: tool
+tool: zfs
+contributors:
+    - ["sarlalian", "http://github.com/sarlalian"]
+    - ["A1EF", "https://github.com/A1EF"]
+filename: LearnZfs-ru.txt
+translators:
+    - ["A1EF", "https://github.com/A1EF"]
+lang: ru-ru
+---
+
+
+[ZFS](http://open-zfs.org/wiki/Main_Page)
+представляет собой переосмысление системы хранения данных, комбинирующее в едином инструменте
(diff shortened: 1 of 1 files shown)


vendethiel:
Hi, I have the several remarks provided by an external reviewer. Can you please look at some? I can easily believe some (most?) are inherited from the document you provided, so they'll need to be fixed there too:

```
this document invented "stripe" 
the whole Виртуальные устройства sexion probably needs to be re-done de novo
and summarise zpoolconcepts(7)
and put it after Пулы хранения and link it as "в пуле, данные страйповане между VDEVами первого уровня:" 
"ZFS датасеты" probably wants to note "транспарентное сжатие"
"Пулы хранения" example 1 "типа raidz из одном устройством для четности"
these look like freebsd examples (by device naming) 
and i think the zpool list -v output doesn't match the input data 
actually none of them do, they're all for a pool created as zpool create zroot gptid/....
the zpool set comment example should note that it's ASCII-only explicitly 
the dataset examples all use tank instead of zroot too 
global replace to match the name in the first sexion
and the mount | grep examples are also freebsd format 
L232 "# Получение всех свойств" example has initial garbage? 
(looks like |123 linenos)
"Установка значения свойства на датасете" must spec compression=zstd
gzip-9 is genuinely and actively malicious 
the first property of snapshots that needs to be mentioned is what they fucking are 
not waffling about the space taken
```

A1EF:
@vendethiel thanks for you reply. I agree with all this remarks. But all of them relative with origin file. My PR is just translation for it. I think I must stand with original until it changed.

vendethiel:
I merged your other PR, thanks.

A1EF:
I adjusted my commit with new changes.

vendethiel:
Thanks :)

PR #4166, opened 2021-05-19
Rename OS X to macOS

verhovsky:
which has been its official name since 2016.


verhovsky:
I did not change all instances that I found because I don't speak all languages and don't know if a conjugation might need to be changed or something.

Menelion (approved):

vendethiel:
Thanks!

PR #4201, opened 2021-07-26
[c/zh-cn] fix typo

mofelee:
image

It confuse me when I learn C

image


mofelee:
I've find a link to explain asterisk for declare pointer.

https://stackoverflow.com/questions/2660633/declaring-pointers-asterisk-on-the-left-or-right-of-the-space-between-the-type

Thoes syntax are same effect, but I prefer to using `int *px` to explain why `not_a_pointer` is not a pointer.

PR #4204, opened 2021-07-28
Fix JSON code blocks in TOML

avrong:
It will correctly show structured, formatted JSON to give a better understanding


Menelion (approved):
Thanks.

PR #4305, opened 2022-01-17
[fish/en] Fish markdown syntax fixes - spacing, etc

GeoMaciolek:

---
(A note for anyone browsing quickly: I'm not the original author of this document; this commit just touches a lot of lines)

I've made some changes to this document to clean up some markdown issues, and to bring it more in-line with existing docs. (The main issue was the use of `>` at the beginning of code lines, leading to [the site](https://learnxinyminutes.com/docs/fish/) rendering "fancy quotes" (`“”` instead of `""`) in chunks of code intended to be copy/pasted, as well as a lack of monospacing etc.

I've elected to use the ` ```fish` header for the large code block, as github's markdown supports this. I've left the other code blocks using bare ` ``` ` marks. (These lines have a have leading `>`, equivalent to `$` in other shells; these would confuse the syntax highlighting.)

There are other changes I'd like to make; those will be for another pull request, I think:
- The generated `learn.fish` isn't what one would expect
- The tone of the article is perhaps more casual than some might prefer
- Some inline code examples could be made somewhat more succinct (with regards to explanation text being split across multiple lines, with the code block between them

### DIFF
--- a/fish.html.markdown
+++ b/fish.html.markdown
@@ -3,6 +3,7 @@ category: tool
 tool: fish
 contributors:
     - ["MySurmise", "https://github.com/MySurmise"]
+    - ["Geo Maciolek", "https://github.com/GeoffMaciolek"]
 filename: learn.fish
 ---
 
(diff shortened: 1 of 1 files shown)


vendethiel:
Thanks.

PR #4310, opened 2022-01-23
[python/en] Removed use of 'is' to compare objects with None in Python.

tarsJr:

A warning has been added in [Python 3.8](https://docs.python.org/3.8/whatsnew/3.8.html#changes-in-python-behavior) to warn comparisons between an object and `None` using `is` as the behaviour is not guaranteed by the language spec.


imba-tjd:
I don't think that's what the warning refers to. The warning only applies to strings and numbers literal, because they will never be None. When you want to compare a normal PyObject with None, `is` is preferred, in my opinion.

PR #4382, opened 2022-04-26
[vyper/en] Update vyper.html.markdown

htadashi:
GitHub now supports Vyper syntax: https://github.com/github/linguist/pull/5778


vendethiel:
We're so far behind with our syntax highlighting that it's not even funny anymore...

PR #4409, opened 2022-06-19
[python/en] Python Update `open` File Open `mode`

TanayParikh:
Per https://docs.python.org/3.7/library/functions.html#open, the `+` is for `open a disk file for updating (reading and writing)`. Per the example, if we're looking to just read/write, the `+` isn't necessary.

### DIFF
--- a/python.html.markdown
+++ b/python.html.markdown
@@ -492,19 +492,19 @@ with open("myfile.txt") as f:
 
 # Writing to a file
 contents = {"aa": 12, "bb": 21}
-with open("myfile1.txt", "w+") as file:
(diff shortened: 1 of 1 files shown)


bowmanjd:
While this does seem like a minor change... I had the same reaction when reading the python/en article the first time. Adding the `+` without there being an obvious reason for it is a bit confusing and uncommon in Python. For what it's worth, I think the change is a worthwhile one.

bowmanjd (approved):
Seems like a minor but worthwhile change to me, given that the code is either reading or writing, not both.

PR #4460, opened 2022-07-21
[lua/en] Update LuaShortReference link

Pab0:
The previous PDF location 404's on lua-users.org, so linking to the respective lua-users.org/wiki page instead (which includes working links to the PDFs on Thomas Lauer's website).

### DIFF
--- a/lua.html.markdown
+++ b/lua.html.markdown
@@ -406,7 +406,7 @@ I started with <a href="https://ebens.me/post/lua-for-programmers-part-1/">Black
 Next I read the official <a href="http://www.lua.org/pil/contents.html">Programming in Lua</a> book.
 That's the how.
 
-It might be helpful to check out the <a href="http://lua-users.org/files/wiki_insecure/users/thomasl/luarefv51.pdf">Lua short
+It might be helpful to check out the <a href="http://lua-users.org/wiki/LuaShortReference">Lua short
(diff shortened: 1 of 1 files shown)


mribeirodantas:
Does this [new link](http://lua-users.org/wiki/LuaShortReference) work for you? In here (Safari 15.5, and Google Chrome 103.0.50), it only works if I set HTTPS in the URL and then there is an issue with the certificate.

Pab0:
> Does this [new link](http://lua-users.org/wiki/LuaShortReference) work for you? In here (Safari 15.5, and Google Chrome 103.0.50), it only works if I set HTTPS in the URL and then there is an issue with the certificate.

On my Ubuntu desktop's browsers (**Firefox** 102.0, **Chromium** 103.0.5060.114, **Chrome** 103.0.5060.134) the plain HTTP site works; when requesting the HTTPS site I get the issue with the certificate as well.
On Android, **Chrome** 103.0.5060 has the same behaviour (allowing HTTP, certificate issue with HTTPS), while **Firefox** (Focus 102.2.0) allows the HTTP site after displaying a warning, and again has the certificate issue with HTTPS.

I thought of going with the HTTP site, since I don't know when the certificates would be renewed.

mribeirodantas:
Ok, then :-)

Thanks.

PR #4540, opened 2022-11-11
[javascript/fr-fr] Fix equality comparison example

kinoute:

The French translation added a paragraph to explain the difference between the `==` and `===` operators. However, in this paragraph, `===` was used twice. The example for `==` was using the wrong operator.


vendethiel:
Merci

PR #4558, opened 2022-12-01
fixed issue json.laod

pranesh467:
python3 json.load() is fixes

### DIFF
--- a/python.html.markdown
+++ b/python.html.markdown
@@ -512,6 +512,7 @@ with open('myfile1.txt', "r+") as file:
 print(contents)
 # print: {"aa": 12, "bb": 21}
 
+import json #import json to load the file content..
 with open('myfile2.txt', "r+") as file:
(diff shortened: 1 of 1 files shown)


bowmanjd (commented):
There is already an `import json` on line 507, making this change redundant. You may be assuming that the lines of code pertaining to a specific concept need to be self-sufficient. 

I suggest closing this PR.

mribeirodantas:
I agree with @bowmanjd. Besides, the [content is targeted at programmers with some experience](https://github.com/adambard/learnxinyminutes-docs/blob/master/CONTRIBUTING.markdown#style-guidelines), so I think if they're trying to run a random snippet and it throws an error due to the library not being imported before, they can solve it. Making sure every snippet of code is self-sufficient could unnecessarily increase the length of the documents, without bringing much benefit for this target audience.

PR #4563, opened 2022-12-13
[vim/en] Replace Command Mode with Normal Mode

conor-f:
This seems a little pedantic, but calling it Normal Mode makes it easier to search online for issues you're having.

### DIFF
--- a/vim.html.markdown
+++ b/vim.html.markdown
@@ -101,9 +101,9 @@ that aims to make getting started with vim more approachable!
 Vim is based on the concept on **modes**.
 
 - Normal Mode - vim starts up in this mode, used to navigate and write commands
-- Insert Mode  - used to make changes in your file
-- Visual Mode  - used to highlight text and do operations to them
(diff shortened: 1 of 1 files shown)


bowmanjd:
I like the change, but after doing some more research, I admit I am uncertain if it is necessary. Take, for instance, [Neovim's docs on the subject](https://neovim.io/doc/user/intro.html#vim-modes). They state:

> In Normal mode you can enter all the normal editor commands.  If you start the editor you are in this mode.  This is also known as command mode.

So... you are right, it is called normal mode. Which is also synonymous with command mode. Thoughts?

conor-f:
My only issue with it is that command mode is very similar to the command line mode which is a separate thing. I'm not sure of the specifics between Vim/Neovim on the naming either.

It's quite a minor change anyways, so feel free to close it if you prefer :)

bowmanjd:
> My only issue with it is that command mode is very similar to the command line mode which is a separate thing.

You have a good point there...

PR #4566, opened 2022-12-19
Using HTML tag to add images

amin13afa:
Using an HTML tag might be the simpler way to add an image to a markdown file.

- [ ] I solemnly swear that this is all original content of which I am the original author
- [ ] Pull request title is prepended with `[language/lang-code]` (example `[python/fr-fr]` or `[java/en]`)
- [ ] Pull request touches only one file (or a set of logically related files with similar changes made)

### DIFF
--- a/markdown.html.markdown
+++ b/markdown.html.markdown
@@ -35,6 +35,7 @@ specific to a certain parser.
 - [Markdownlint](#markdownlint)
 - [Further reading](#further-reading)
 
+
 ## HTML Elements
 
 Markdown is a superset of HTML, so any HTML file is valid Markdown.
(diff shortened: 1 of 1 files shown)


bowmanjd:
I like that this is a practical example of the already stated "Markdown is a superset of HTML" but I don't understand how using HTML is "the simpler way" than just using Markdown. Is there something to be gained by suggesting to a markdown newcomer that they should use the HTML image tag instead of markdown in this instance?

Also concerned that the width and height are specified, outside CSS, which may have unintended consequences for, say, static site generators and the like.

If you are eager to see this change, @amin13afa , you may wish to edit your PR title and comments to comply with contributing guidelines. Check the boxes, follow the second point, etc.

PR #4598, opened 2023-02-16
[typescript/en] Expanding Enums and Alternatives

Buttars:
This addition expands on enums and mentions the pitfalls enums may have and offers alternatives. This is mentioned explicitly in the typescript docs since the post transpiration behavior of enums can be inconsistent.

The typescript documentation recommends using const enums or const objects as alternatives. For most use cases these are preferable since they offer a more consistent run time result.

### DIFF
--- a/typescript.html.markdown
+++ b/typescript.html.markdown
@@ -1,15 +1,15 @@
 ---
 language: TypeScript
 contributors:
-    - ["Philippe Vlérick", "https://github.com/pvlerick"]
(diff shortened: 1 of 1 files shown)


verhovsky:
Enums are already mentioned, I don't think we need this much about enums.

Buttars:
Would you be open to documentation about the pitfalls of enums? Specifically the const as const alternative pattern without the additional enum explanations? Enums are a particularly controversial feature due to their runtime inconsistencies and I believe it makes sense to show the alternative in this type of documentation.

PR #4609, opened 2023-03-03
[sql/en] More database-agnostic SQL

### LINKED ISSUE #4602: [sql/en] LIMIT is non-ANSI and not universally supported
The SQL page is succinct and well designed, but I worry about the `LIMIT` clause. While supported by MySQL, MariaDB, and PostgreSQL, and SQLite, `LIMIT` is not standard ANSI SQL, and is notably not supported by Microsoft SQL Server (T-SQL).

I could submit a PR if appropriate, but would like to know if it is permissible to show two different options, for different database engines.

For instance, we could start with

`SELECT * FROM departments ORDER BY dept_name OFFSET 0 ROWS FETCH NEXT 5 ROWS ONLY;`

which is ANSI compliant and works on MariaDB, PostgreSQL, and SQL Server, then follow with something like

> Some database engines such as MySQL And SQLite do not support ORDER BY... OFFSET... FETCH... and instead implement a LIMIT clause:
>
> `SELECT * FROM departments LIMIT 5`

In related news, perhaps the whole article could lean harder toward PostgreSQL or MariaDB for compliance's sake? For instance, we could use the [Chinook sample data](https://github.com/lerocha/chinook-database/tree/master/ChinookDatabase/DataSources), instead of a sample with MySQL-only instructions.

At the end of the day, I just really want to recommend learnxinyminutes to anyone, regardless of database engine.

bowmanjd:

Fixes #4602 
A lot of changes on this one that I hope promote discussion:
- I tested my changes on MySQL, MariaDB, PostgreSQL, Microsoft SQL Server, SQLite, IBM DB2, and Oracle. If a statement doesn't work or works differently on one of these engines, I noted it.
- Instead of referencing the MySQL-specific employees database, I introduced CREATE and INSERT statements that populate a minimalist data sample for querying later, and altered all subsequent queries accordingly.
- I favored ANSI and then explain variants where appropriate, such as with ORDER BY... OFFSET... FETCH... vs LIMIT vs TOP
- I added a link to the Sakila database which is a fuller option and supports a wide range of engines

Questions:
- How best to show variants, such as LIMIT or TOP? Should they be commented or left as-is? They will not run on certain engines.
- Is the created-on-the-fly sample tables a good idea, or is it better to just use something like the Sakila sample database throughout
- Should we be describing the CREATE DATABASE and USE syntax at all, since it varies so much across engines?


bowmanjd:
Would it be more helpful if I broke these changes out into multiple PRs? Happy to do so if anything here is helpful.

verhovsky:
Sorry but this makes the article longer for no real benefit for most people. The link to http://troels.arvin.dk/db/rdbms should be enough for the people who need to know about it.

PR #4704, opened 2023-07-10
[c/en] Wrapping fixes and some copy edits/small additions

sjrct:
This does a few things with the c/en tutorial.

1) Fix the word wrapping. A lot of lines were longer than 80 characters. Some small copy edits here to prevent line breaks in strange places.
2) Added note about `void` in to signal empty parameter lists and add `void` to existing empty param functions
3) Note about not using braces for one-statement control blocks.
4) Note about using field initializers for structs
5) Note about `#pragma once`, and about `static inline` funcs in headers. Wasn't sure if this was too niche to be usefully included.
6) Small copy edits

---

### DIFF
--- a/c.html.markdown
+++ b/c.html.markdown
@@ -371,8 +371,8 @@ int main (int argc, char** argv)
   printf("\n");
 
   // *****NOTES*****:
-  // If the body of control expression is only one statement than parenthesises
-  // are optional, but often recommended:
+  // If the body of control expression is only one statement then parentheses
+  // are optional, but recommended:
   for (jj = 0; jj < 5; jj++)
       printf("%d\n", jj);
 
--- a/c.html.markdown
+++ b/c.html.markdown
@@ -27,7 +27,8 @@ memory management and C will take you as far as you need to go.
 >
 > `-Wall -Wextra -Werror -O2 -std=c99 -pedantic`
 >
-> For information on what these flags do as well as other flags, consult the man page for your C compiler (e.g. `man 1 gcc`) or just search online.
+> For information on what these flags do as well as other flags, consult the man
+> page for your C compiler (e.g. `man 1 gcc`) or just search online.
 
 ```c
 // Single-line comments start with // - only available in C99 and later.
@@ -63,7 +64,7 @@ enum days {SUN = 1, MON, TUE, WED = 99, THU, FRI, SAT};
 // libraries for the headers.
 // For your own headers, use double quotes instead of angle brackets, and
 // provide the path:
-#include "my_header.h" 		// local file
+#include "my_header.h" // local file
 #include "../my_lib/my_lib_header.h" //relative path
 
 // Declare function signatures in advance in a .h file, or at the top of
@@ -71,18 +72,20 @@ enum days {SUN = 1, MON, TUE, WED = 99, THU, FRI, SAT};
 void function_1();
 int function_2(void);
 
-// At a minimum, you must declare a 'function prototype' before its use in any function.
-// Normally, prototypes are placed at the top of a file before any function definition.
+// At a minimum, you must declare a 'function prototype' before its use in any
+// function. Normally, prototypes are placed at the top of a file before any
+// function definition.
 int add_two_ints(int x1, int x2); // function prototype
-// although `int add_two_ints(int, int);` is also valid (no need to name the args),
-// it is recommended to name arguments in the prototype as well for easier inspection
+// although `int add_two_ints(int, int);` is also valid (no need to name the
+// args), it is recommended to name arguments in the prototype as well for
+// easier inspection
 
 // Function prototypes are not necessary if the function definition comes before
-// any other function that calls that function. However, it's standard practice to
-// always add the function prototype to a header file (*.h) and then #include that
-// file at the top. This prevents any issues where a function might be called
-// before the compiler knows of its existence, while also giving the developer a
-// clean header file to share with the rest of the project.
+// any other function that calls that function. However, it's standard practice
+// to always add the function prototype to a header file (*.h) and then #include
+// that file at the top. This prevents any issues where a function might be
+// called before the compiler knows of its existence, while also giving the
+// developer a clean header file to share with the rest of the project.
 
 // Your program's entry point is a function called "main". The return type can
 // be anything, however most operating systems expect a return type of `int` for
@@ -125,7 +128,8 @@ int main (int argc, char** argv)
   short x_short = 0;
 
   // chars are defined as the smallest addressable unit for a processor.
-  // This is usually 1 byte, but for some systems it can be more (ex. for TMS320 from TI it's 2 bytes).
+  // This is usually 1 byte, but for some systems it can be more (ex. for TMS320
+  // from TI it's 2 bytes).
   char x_char = 0;
   char y_char = 'y'; // Char literals are quoted with ''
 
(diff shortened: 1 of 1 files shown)


c.html.markdown
```diff
@@ -366,13 +371,17 @@ int main (int argc, char** argv)
...
+      printf("%d\n", jj);
+
+  // Loops and if statements MUST have a body. If no body is needed:
   int i;
   for (i = 0; i <= 5; i++) {
-    ; // use semicolon to act as the body (null statement)
   }
   // Or
-  for (i = 0; i <= 5; i++);
+  for (i = 0; i <= 5; i++); // use semicolon to act as the body (null statement)
```
verhovsky, line 384: Moving this comment is wrong, the `;` is a statement terminator on this line, not a body.

sjrct: Every `for` loop has four parts, one of which is the statement to be executed (aka the body of the for loop). The statement could either be a simple statement, such as a call to `printf` as above, a compound statement (aka block) that is a collection of statements enclosed in `{` and `}`, or the null statement, which is a statement that just doesn't do anything, which is just written as `;`. See [this section of the gnu C manual](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#The-Null-Statement) for reference.

The previous way it was written, it was a null statement inside of a compound statement and the comment implied that the semicolon inside the curlies was necessary when it wasn't. In real life C, for loops that do nothing will have either `;` or `{}`. Although `{;}` as written here would've worked, I've never personally seen it and I don't think it is intuitive.

I certainly think what's going on could be better explained, but I also think that the null statement isn't something that is important enough to spend anything more than a passing mention (if that) on in this introduction to C.

c.html.markdown
```diff
@@ -541,19 +552,19 @@ int main (int argc, char** argv)
...
-  printf("%d\n", *(my_ptr + 21)); // => Prints who-knows-what? It may even crash.
+  // Dereferencing memory that you haven't allocated gives "unpredictable
+  // results" - the program is said to invoke "undefined behavior"
+  printf("%d\n", *(my_ptr + 21)); // => Prints who-knows-what? It may even crash
 
   // When you're done with a malloc'd block of memory, you need to free it,
-  // or else no one else can use it until your program terminates
-  // (this is called a "memory leak"):
+  // or else no one else can use it until your program terminates (this is
+  // called a "memory leak"):
```
verhovsky, line 561: If it's gonna be on two lines you might as well keep it

```suggestion
  // or else no one else can use it until your program terminates
  // (this is called a "memory leak"):
```

c.html.markdown
```diff
@@ -541,19 +552,19 @@ int main (int argc, char** argv)
...
   free(my_ptr);
 
   // Strings are arrays of char, but they are usually represented as a
   // pointer-to-char (which is a pointer to the first element of the array).
-  // It's good practice to use `const char *' when referring to a string literal,
-  // since string literals shall not be modified (i.e. "foo"[0] = 'a' is ILLEGAL.)
+  // It's good practice to use `const char *' when referring to a string literal
+  // since string literals shall not be modified (i.e. foo[0] = 'a' is ILLEGAL.)
```
verhovsky, line 567: this is definitely clearer

```suggestion
  // since string literals shall not be modified (i.e. "foo"[0] = 'a' is ILLEGAL.)
```

c.html.markdown
```diff
@@ -864,11 +894,16 @@ as the C file.
...
+/* happens in the case of circlular dependencies, or such as when a header is */
+/* included alongside a header that includes the same header itself.          */
 #ifndef EXAMPLE_H /* if EXAMPLE_H is not yet defined. */
 #define EXAMPLE_H /* Define the macro EXAMPLE_H. */
 
+/* This is a less verbose non-standard alternative to the include guards just */
+/* mentioned. While non-standard it avoids errors from the two header files   */
+/* accidentally defining the same macro. Place it at the top of the header.   */
+#pragma once
+
```
verhovsky, line 906: non standard, shouldn't be mentioned

```suggestion
```

c.html.markdown
```diff
@@ -366,13 +371,17 @@ int main (int argc, char** argv)
   printf("\n");
 
   // *****NOTES*****:
-  // Loops and Functions MUST have a body. If no body is needed:
+  // If the body of control expression is only one statement than parenthesises
+  // are optional, but often recommended:
```
verhovsky, line 375:
```suggestion
  // If the body of control expression is only one statement then parentheses
  // are optional, but recommended:
```

c.html.markdown
```diff
@@ -63,26 +65,28 @@ enum days {SUN = 1, MON, TUE, WED = 99, THU, FRI, SAT};
...
-#include "my_header.h" 		// local file
+#include "my_header.h" // local file
 #include "../my_lib/my_lib_header.h" //relative path
 
 // Declare function signatures in advance in a .h file, or at the top of
 // your .c file.
-void function_1();
-int function_2(void);
+void function_1(void);
+int function_2(int a, float b);
```
verhovsky, line 74: I think it's fine the way it is and is easier to understand when it's simpler like this. I understand `()` is a compiler warning but in practice it doesn't matter if you don't even know how to access the variable length arguments of functions in C. I think mentioning `(void)` at the end is good but keeping it simple at the beginning is better.

```suggestion
void function_1();
int function_2(void);
```

c.html.markdown
```diff
@@ -125,7 +129,8 @@ int main (int argc, char** argv)
   short x_short = 0;
 
   // chars are defined as the smallest addressable unit for a processor.
-  // This is usually 1 byte, but for some systems it can be more (ex. for TMS320 from TI it's 2 bytes).
+  // This is usually 1 byte, but for some systems it can be more (ex. for TMS320
+  // from TI it's 2 bytes).
```
verhovsky, line 133:
```suggestion
  // This is usually 1 byte, but for some systems it can be more
  // (ex. for TMS320 from TI it's 2 bytes).
```

c.html.markdown
verhovsky:
```suggestion
  for (i = 0; i <= 5; i++); // semicolon acts as the body (null statement)
```

c.html.markdown
```diff
@@ -672,15 +683,15 @@ printIntArray(my_arr, size);
 // will print "arr[0] is: 1" etc
 */
 
-// if referring to external variables outside function, you should use the extern keyword.
+// if referring to external variables outside function, use the extern keyword.
 int i = 0;
-void testFunc() {
+void testFunc(void) {
   extern int i; //i here is now using external variable i
```
verhovsky, line 689:
```suggestion
void testFunc() {
  extern int i; // i here is now using external variable i
```

c.html.markdown
```diff
@@ -672,15 +683,15 @@ printIntArray(my_arr, size);
...
 }
 
 // make external variables private to source file with static:
 static int j = 0; //other files using testFunc2() cannot access variable j
-void testFunc2() {
+void testFunc2(void) {
```
verhovsky, line 694:
```suggestion
void testFunc2() {
```

c.html.markdown
```diff
@@ -693,6 +704,18 @@ void testFunc2() {
 // declared with some other starting value.
 //**You may also declare functions as static to make them private**
 
+// NOTE that before C23, and unlike C++, functions taking no arguments without
```
verhovsky, line 707:
```suggestion
// Note that before C23, and unlike C++, functions taking no arguments without
```

c.html.markdown
sjrct: Ifrc correctly lines longer than 80 characters were causing some rendering issues

c.html.markdown
sjrct: Even with `-Wall -Wextra` gcc (at least version 13.2.0 for me) compiles calling functions with `()` for the parameter list with some arbitrary parameters without any warning. clang does give me some warning. C23 did change this behavior though, making `(void)` and `()` equivalent parameter lists for function declarations, and I know gcc is on v14 now, so my OS's version is probably a bit out of date and this will not become as relevant for new C in not too long.

Having been burning by that unexpected behavior in the past, it was my intuition to include it when making these changes originally, but with the C23 changes putting it off to the end makes sense, or even leaving it out. I'll make some changes shortly

c.html.markdown
sjrct: I think this change makes sense but I also think that indexing a string literal will probably make readers take a double take, and I wonder if we could just drop the parenthetical entirely.

PR #4705, opened 2023-07-10
[make/en] Add info on custom functions & auto generating dependency rules

sjrct:
Adds a bit about custom functions and a section about automatically generating dependency rules

---


verhovsky:
I checked a few Makefiles I have (curl's and bash's) and they don't use `call`. The new section about generated dependency rules I didn't really understand and it seems pretty specific, I don't think it would be helpful in this guide.

sjrct:
Fair enough, although not sure those are good examples since they use generated makefiles. I suppose I assumed the target audience to be those who would be writing makefiles rather than reading generated ones

PR #4760, opened 2023-09-22
Update toml.html.markdown

KronosTheLate:
IMPORTANT: I have only updated the introductory text. I do not know what changes have happened between 0.4 and 1.0, so the contents of the file may be different. I would like help in updating the actual examples. It seems like copying the snippets found on [this page](https://toml.io/en/v1.0.0) would be a good starting point.


verhovsky:
#4690

Issue #10, opened 2013-06-28
Add further reading and topics not covered for python


irfancharania:
I'm new to python and I think it'd be really helpful to see a "further reading" and "topics/standard libraries not covered" section like in lua and others

PR #4819, opened 2024-01-01
[Tailspin/en] Add page for the Tailspin language

tobega:

### DIFF
--- a/tailspin.html.markdown
+++ b/tailspin.html.markdown
@@ -1,0 +1,335 @@
+---
+language: tailspin
+filename: learntailspin.tt
+contributors:
+    - ["Torbjörn Gannholm", "https://github.com/tobega/"]
+
+---
+
+**Tailspin** works with streams of values in pipelines. You may often feel
+that your program is the machine and that the input data is the program.
+
+While Tailspin is unlikely to become mainstream, or even production-ready,
+it will change the way you think about programming in a good way.
+
+```tailspin
+// Comment to end of line
+
+// Process data in a pipeline
+'Hello, World!' -> !OUT::write
+
+// Output newlines:
+'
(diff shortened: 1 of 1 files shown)


vendethiel:
Honestly, I think that while this might be useful as a syntax reference, it's little more for someone who hasn't already read about the language.
It's missing a lot of higher-level explanations, and from the very beginning:
- doesn't explain that `->` is the basic operator to "send" data in a pipeline
- doesn't explain what `!` means and where it can appear
- doesn't explain that `'` is special
- doesn't explain what `OUT::write` means
- doesn't explain what `$#10;` means
- doesn't explain matcher syntax (inside of `<>`, what `?()` means, what `~()` would mean)
- doesn't explain what `~/` does

I think if this it to be a learning resource, it needs at least this.

tobega:
Thanks, good feedaback! I guess I went too far on "Prefer example to exposition"

PTAL

vendethiel:
Thank you, that's great.

> Thanks, good feedaback!

I actually went and read a few blog posts explaining how the language works etc before it clicked.

> I guess I went too far on "Prefer example to exposition"

I think the more you deviate from what most people are used to, the more you'll have to explain bits and pieces.
Especially when your syntax is symbol-heavy, it can be different to wrap your head around.
I've learned most of the articles that are here - though I'd never pretend to actually know a tenth of them - and I've used a few languages but it can still take some time getting used totally novel syntax.

vendethiel:
Thank you :).

tobega:
Thanks! It seems to give a 404 on the actual site. Did I do something wrong or does it just take a while?

vendethiel:
Ah, our pygments is way outdated

PR #4887, opened 2024-04-07
[CPP/en] Added content for string length

Git-Harshit:
Updated string length support from C++ using `string::length()` and `string::size()` from the included <strings> header file and additionally including <cstrings> header to show finding length for C-compliant char* pointer or char[] array type strings with `strlen()`.

# Checklist
- [X] I solemnly swear that this is all original content of which I am the original author
- [X] Pull request title is prepended with `[language/lang-code]` (example `[python/fr-fr]` or `[java/en]`)
- [X] Pull request touches only one file (or a set of logically related files with similar changes made)
- [ ] Content changes are aimed at *intermediate to experienced programmers* (this is a poor format for explaining fundamental programming concepts)
- [ ] If you've changed any part of the YAML Frontmatter, make sure it is formatted according to [CONTRIBUTING.md](https://github.com/adambard/learnxinyminutes-docs/blob/master/CONTRIBUTING.markdown)
  - [X] Yes, I have double-checked quotes and field names!

PR #4888, opened 2024-04-07
[cpp/en] String length content addition

Git-Harshit:
Updated `string::length()` and `string::size()` usage demonstration for finding the length for individual strings.
Included <cstring> header to show `strlen()` usage for char* pointer or char[] array based C-compliant string length.
Got my #4887 closed due to forked repository deletion, so re-creating a PR to propose the same changes.

### DIFF
--- a/c++.html.markdown
+++ b/c++.html.markdown
@@ -228,6 +228,9 @@ cout << myString + myOtherString; // "Hello World"
 
 cout << myString + " You"; // "Hello You"
 
+// C++ string length can be found from either string::length() or string::size()
+cout << myString.length() + myOtherString.size(); // Outputs 11 (= 5 + 6).
+
 // C++ strings are mutable.
 myString.append(" Dog");
 cout << myString; // "Hello Dog"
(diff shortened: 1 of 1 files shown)


c++.html.markdown
```diff
@@ -228,10 +228,20 @@ cout << myString + myOtherString; // "Hello World"
...
+cout << myString.length() + myOtherString.size();
+
 // C++ strings are mutable.
 myString.append(" Dog");
 cout << myString; // "Hello Dog"
 
+// C++ can handle C-style strings with related functions using cstrings
+#include <cstring>
+
+char myOldString[20] = "Hello CPP";
```
vendethiel, line 242: Why 20?

Git-Harshit: > Why 20?

No specific reason, just initialized with a sufficient size while I was choosing on string content. Shall I set it to around 9-10 for a better fit?

vendethiel: 10 for the null, otherwise your example won't work. Or just leave the length empty...

Git-Harshit: > 10 for the null, otherwise your example won't work. Or just leave the length empty...

Right. Thanks for suggesting, I'll update it to 10

vendethiel:
thanks

PR #4607, opened 2023-03-01
[gdscript/en] Update for Godot 4

zacryol:
resolves #4470 

- [x] I solemnly swear that this is all original content of which I am the original author
  - (It's based on the original GDScript section, so not entirely mine)
- [x] Pull request title is prepended with `[language/lang-code]` (example `[python/fr-fr]` or `[java/en]`)
- [x] Pull request touches only one file (or a set of logically related files with similar changes made)
- [x] Content changes are aimed at *intermediate to experienced programmers* (this is a poor format for explaining fundamental programming concepts)
- [x] If you've changed any part of the YAML Frontmatter, make sure it is formatted according to [CONTRIBUTING.md](https://github.com/adambard/learnxinyminutes-docs/blob/master/CONTRIBUTING.markdown)
  - [x] Yes, I have double-checked quotes and field names!

Should cover:
 - [x] use of `@onready`
 - [x] use of various `@export` annotations
   - [x] base `@export`
   - [x] ~~more specific annotations~~ Probably not needed for a basic introduction.
 - [x] Documentation Comments
 - [x] enums as types
 - [x] Packed Arrays instead of Pool
 - [x] New NodePath syntax (`^` instead of `@`)
 - [x] Directly using a Node as an export type
 - [x] Remove the use of a Tween node
 - [x] Typed signals
 - [x] Typed Arrays
 - [x] Lambda expressions
 - [x] More emphasis on static typing (more of an opinion)
   - Added stuff to the section
 - [x] Change superclass method call to `super()`
 - [ ] Anything else that people may bring up.

PR #4890, opened 2024-04-09
[gdscript/*] update NodePath literal syntax

verhovsky:
@Wichamir it seems like this was changed in GDScript 4

https://docs.godotengine.org/en/3.0/classes/class_nodepath.html#description

https://docs.godotengine.org/en/4.0/classes/class_nodepath.html#description

Could you read over your page again, maybe there are other things that need updating?


Wichamir:
@verhovsky This whole tutorial is for Godot 3 and thus for old GDScript, so if you want to introduce such changes, then the whole text should be updated to fit Godot 4. Actually there is already a PR for this - #4607.

PR #4893, opened 2024-04-16
[lean4/en] Documentation for Lean 4.

balu:

This documentation contains more comments than the usual tutorials in this repository. Most of it explains theorem proving and not fundamental programming concepts. I think it is still aimed at intermediate to experienced programmers. Please let me know if I should change it.

### DIFF
--- a/lean4.html.markdown
+++ b/lean4.html.markdown
@@ -10,13 +10,6 @@ language and an interactive theorem prover.
 
 ```lean4
 /-
-Imports have to be at the top of the file.
-
-We will need `use` for some of the proofs below.
--/
-import Mathlib.Tactic.Use
-
-/-
 An enumerated data type.
 -/
 inductive Grade where
@@ -77,7 +70,7 @@ defined for that type.
 Using this we can prove `1 = 1` as follows.
 -/
 
-def one_eq_one : 1 = 1 := Eq.refl 1
+theorem one_eq_one : 1 = 1 := Eq.refl 1
 
 /-
 But there is no way to prove (construct a value of type) `0 = 1`.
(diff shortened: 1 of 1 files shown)


lean4.html.markdown
```diff
@@ -0,0 +1,587 @@
...
+/-
+`Exists.intro v h` proves `∃ x, p x` by substituting `x` by `v` and using the
+proof `h` for `p v`.
+-/
+
+/-
+Now, we will see how to use hypothesis that are existentials to prove
+conclusions that are existentials.
+-/
+theorem even_mul_even_is_even' (hn : Even n) (hm : Even m) : Even (n*m) :=
```
Ferinko, line 218: One should probably bind the `m` and `n` and not rely on Lean's autoimplicit in this expository piece.
Something like `theorem even_mul_even_is_even' {m n} ...`, or with types, or with an explicit `∀ {m n}`.

Ferinko: This is probably all around the doc.

lean4.html.markdown
```diff
@@ -0,0 +1,587 @@
...
+theorem  succ_of_odd_is_even : Odd n → Even (n+1) := sorry
+
+/-
+We can use implications by applying them.
+-/
+example : Odd 1 := by
+  apply succ_of_even_is_odd
+  exact zero_even
+/-
+The two new tactics are:
```
Ferinko, line 285: `use` is a stronger apply (it's a `refine`) and has been used harum scarum. Might deserve the same treatment as the following two.

lean4.html.markdown
```diff
@@ -0,0 +1,587 @@
...
+
+  - `apply p` where `p` is an implication `q → r` and `r` is the goal rewrites
```
Ferinko, line 287: This is not quite true. e.g. `example : True := by apply True.intro` is fine even tho `True.intro` is not an implication. It unifies the conclusion and generates a goal for every hypothesis.

lean4.html.markdown
```diff
@@ -0,0 +1,587 @@
...
+-/
+theorem even_or_odd (n : Nat) : Even n ∨ Odd n := by
+  induction n
+  case zero => left ; exact zero_even
+  case succ n ihn =>
+    cases ihn with
+    | inl h => right ; apply (succ_of_even_is_odd h)
+    | inr h => left  ; apply (succ_of_odd_is_even h)
+/-
+`induction` is not just for natural numbers. It is for any inductive type.
```
Ferinko, line 327: And all types are inductive.
```example (h : True) : True = True := by
  induction h -- ok
  sorry```

File: lean4.html.markdown [hunk 5]
```diff
@@ -0,0 +1,587 @@
...
+  case inr h => specialize hnqnp h ; contradiction
+
+/-
+To illustrate how we can define an work with axiomatic systems. Here is a
+definition of Groups and some proofs directly translated from "Topics in
+Algebra" by Herstein, Second edition.
+-/
+
+/-
+A class defines a set of axioms
```
Ferinko, line 440: This is misleading.
`class A where a : Int` is perfectly valid. (Yes, this can be viewed as an axiom, just don't do it.)

lean4.html.markdown
```diff
@@ -0,0 +1,587 @@
...
+/-
+We prove that the identity element is unique.
+-/
+theorem identity_element_unique : ∀ e' : G, is_identity e' → e' = e := by
+  intro e'
+  intro h
+  specialize h e
+  obtain ⟨h1, _⟩ := h
+  have h' := identity e'
+  obtain ⟨_, h2⟩ := h'
```
Ferinko, line 472: Why have we started using `obtain` instead of the `let` up until now?

lean4.html.markdown
```diff
@@ -0,0 +1,587 @@
...
+  apply Eq.trans (Eq.symm h2) h1
```
Ferinko, line 473: This should probably say `exact` now that you introduced it above.

lean4.html.markdown
```diff
@@ -0,0 +1,587 @@
...
+    _ = ai * a * x   := by rw [h2]
+    _ = ai * (a * x) := by rw [assoc]
+    _ = ai * (a * y) := by rw [h3]
+    _ = ai * a * y   := by rw [← assoc]
+    _ = (e : G) * y  := by rw [h2]
+    _ = y            := (identity y).right
+
+/-
+Mutually recursive definitions have to be carefully constructed. In particular,
+the hypotheses when recursing should imply well-foundedness.
```
Ferinko, line 499: This is misleading. Every definition in Lean has to be well-founded, regardless of whether it's mutually defined.

lean4.html.markdown
```diff
@@ -0,0 +1,587 @@
...
+    intro ; contradiction
+  case succ =>
+    constructor
+    intro a ; have := a.right ; contradiction
+    intro a ; have := a.right ; contradiction
+
+/-
+Programming algorithms in Lean is harder.
+
+This is because we have to convince Lean of termination. Here's a simple
```
Ferinko, line 532: This is not true. Lean accepts `partial` definitions that can be computed with just fine; just can't be unfolded to reason with.

lean4.html.markdown
```diff
@@ -0,0 +1,587 @@
...
+      @go (Nat.succ d) n (by omega)
+  termination_by (n - d)
+/-
+We have to specify that the recursive function `go` terminates because `n-k`
+decreases in each recursive call. This needs the hypothesis `n > k` at the
+recursive call site. But the function itself can only assume that `n ≥ k`. We
+label the test `n ≤ k` by `h` so that the falsification of this proposition can
+be used by `omega` later to conclude that `n > k`.
+
+The tactic `omega` is a general-purpose solver for equalities and inequalities.
```
Ferinko, line 556: This is also not true. `omega` is a specific-purpose decision procedure for deciding problems in presburger arithmetic.

Ferinko (changes requested):

lean4.html.markdown
```diff
@@ -0,0 +1,587 @@
...
+
+/- def zero_eq_one : 0 = 1 := Eq.refl 0 -/
+
+/-
+Let us prove a more complicated equality.
+
+The following will fail because both sides are not "obviously equal".
+-/
+
+/- def pf : 2^9-2^8 = 2^8 := Eq.refl (2^8) -/
```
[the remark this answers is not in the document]
balu: I have removed this theorem altogether. You are right, this theorem does not make sense to me after reading your comments.

lean4.html.markdown
balu: Are `obtain` and `let` interchangeable?

Ferinko: They're not. `obtain` is a `have` + full expressivity of what you call 'rcases patterns'. `let` is a proof-relevant binder that just happens to be able to deconstruct some simple patterns.

E.g. the former can do things like:
```have X : True ∨ False := sorry
  obtain h₁ | h₂ := X```
  
I would say `obtain` is almost a superset but that's somewhat misleading because `let's` are meant to be used for a different purpose + there's the distinction of whether they keep the data on their rhs around.

So, just think of them as two different features. `obtain` allows you to inspect terms, `let` allows you to introduce new terms.

balu: I see. I have replaced all `let` with `have` in proofs. I think in proofs this is the right thing to do due to proof irrelevance as you mentioned.

Thanks for your comments. They have been very helpful.

verhovsky:
@Ferinko I'm merging this. If there's other issues or parts of your review that weren't addressed or any other things you want to add, it would be nice if you submitted a pull request.

PR #4904, opened 2024-04-21
[html/en] Add more tags

Th3G33k:
meta, script, style, div, form, iframe

### DIFF
--- a/html.html.markdown
+++ b/html.html.markdown
@@ -161,6 +161,71 @@ useful tips.
 		<td>second row, second column</td>
 	</tr>
 </table>
+
+<!-- ################ Additional Tags ################# -->
+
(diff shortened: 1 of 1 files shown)


verhovsky:
I don't like how this is adding a third file-within-the-file. LXIYM should just be a single commented file, but I can see the value of having a simple version of the HTML document and then a commented version, but adding a third one doesn't seem right.

Th3G33k:
Thanks for the review.

It is already separated in three parts. The third part starts from here :

```html
<!-- And that's it, creating an HTML file can be simple. -->

<!-- But it is possible to add many additional types of HTML tags. -->

<!-- The <img /> tag is used to insert an image. -->
```

I was trying to add more useful tags.

PR #4909, opened 2024-04-26
[gno/en] Add Gno Tutorial

leohhhn:
Hello, we would like to add the Learn X in Y tutorial for [Gno](https://docs.gno.land). 

- [x] I solemnly swear that this is all original content of which I am the original author
- [x] Pull request title is prepended with `[language/lang-code]` (example `[python/fr-fr]` or `[java/en]`)
- [x] Pull request touches only one file (or a set of logically related files with similar changes made)
- [x] Content changes are aimed at *intermediate to experienced programmers* (this is a poor format for explaining fundamental programming concepts)
- [ ] If you've changed any part of the YAML Frontmatter, make sure it is formatted according to [CONTRIBUTING.md](https://github.com/adambard/learnxinyminutes-docs/blob/master/CONTRIBUTING.markdown)
  - [ ] Yes, I have double-checked quotes and field names!

cc @zivkovicmilos @michelleellen

PR #4912, opened 2024-04-27
Remove cryptocurrency promotion

verhovsky:
#4909 reminded me that I wanted to do this.

These esoteric programming languages are not interesting as programing languages. Having them on the website serves no purpose. We all know their real purpose.

I'm just slightly annoyed when I scroll the list of languages, see one of these and think the name sounds interesting, go to check it out and realize I just wasted some time and brain space on crypto stuff. I imagine other users of the site have a similar experience.

I think [Solidity](https://learnxinyminutes.com/docs/solidity/) can stay since it's kind of a real thing and if someone makes one for Bitcoin's [Script](https://en.bitcoin.it/wiki/Script) that should also be accepted. Vyper arguably as well, but from my googling around it seems that it might've [been abandoned in 2020](https://blog.ethereum.org/2020/01/08/update-on-the-vyper-compiler) so it shouldn't be a big loss.

@adambard @vendethiel


vendethiel:
I tentatively agree, though I haven’t checked which ones are used for what.

danielytics:
I don't understand why Solidity can stay but others can't. Maybe Vyper is abandoned, but FunC and Tact are just as real as Solidity, used actively by TON and telegram mini apps. They're even interesting in their own right as actor-based message passing languages (I certainly find Tact very interesting just from a language design point of view).

I can understand not liking or wanting to promote blockchain, but it does seem weird to restrict what languages you allow based on what their execution target is.

Also, what is "interesting" is arbitrary, just because something is or isn't interesting to you doesn't mean it isn't interesting or useful to somebody else. I personally don't find golfscript or bf interesting. Similarly, I don't find proprietary languages like kdb+ that I can't actually use without paying interesting and would find reading that page like I wasted my time when I later find out I can't actually run what I just read. But I'd never suggest that they should be removed.

If you really don't want to have blockchain languages, that's obviously your call, but at least be consistent about it then.

verhovsky:
Because it's the first and/or most successful Turing complete blockchain language. It's old now so having it feels more of like an honest description of a language that actually exists in the social sphere instead of using this website to try and promote something new into existing.

> I certainly find Tact very interesting just from a language design point of view

Looks like a normal C-like language to me.

> I personally don't find golfscript or bf interesting. Similarly, I don't find proprietary languages like kdb+ that I can't actually use without paying interesting

with the exception of brainfuck, me too.

> be consistent

Why? We can make opinionated editorial decisions. We can just think about things and do what feels reasonable, our behaviour doesn't need to be compressible into some formally specified policy like `if blockchain: delete()`.

danielytics:
> I could hear someone suggesting we add a disclaimer to pages describing proprietary languages.

Couldn't the same be said for blockchain languages? The original complaint was that spending time on a language that looks interesting wastes time and brainspace when later realizing that it was a blockchain language. By stating this upfront, the reader will immediately know whether they should continue reading or not.

> Looks like a normal C-like language to me.

If you are just looking at the syntax, sure, but if that's all that you judge a language on, I feel that Learn X in Y Minutes has a lot of "uninteresting" languages that are just "normal X-like languages".

The semantics matter and Tact is an actor-model message-passing based language. That is, Tact defines messages and for Tact to call other Tact (or FunC, they are interoperable) programs, you send an asynchronous message. Each program is independent and asynchronous, only calls within a single program are transactional. FunC works similarly, but is lower level, Tact makes messages a language-level construct and slo distinguishes between normal (synchronous) functions, message handlers (asynchronous functions that are called on receipt of a message) and getters (side-effect free pure functions that are used to retrieve data from a program but cannot modify any state. I personally find that interesting. If you don't, then I can't really change your mind, but that's kinda my point here, "interesting" is subjective.

> We can make opinionated editorial decisions. 

Ultimately, its not my repo, so I can just give you my opinion and leave it up to you. I personally value consistency in decision making, but if you don't, I won't argue about it.

Riley-Kilgore:
Just had a PR https://github.com/adambard/learnxinyminutes-docs/pull/4999 closed based on this conversation.

@verhovsky I hope that further consideration is made in due time regarding this conversation.

I like the idea that @vendethiel has pointed out, that a disclaimer could be present to denote that a language compiles to a target which may be specific to some execution environment.

There are a growing number of blockchain-related languages, many of which have substantial usage in production settings. As such, 'using this website to try and promote something new into existing' seems highly impractical.

Thank you for the opportunity to express my opinion on the matter.

nielstron:
I am not a frequent user of this and have no strong opinion about having smart contract langs on the website or not.

However I do think the website would in general benefit from a more fine grained categorization of languages. i.e.
- formal theorem prover languages (Coq, lean)
- markup /formatting langs (yaml, latex, markdown,...)
- general purpose, functional, ...

One of these tags could be crypto/smart contract, allowing everyone to filter for interesting langs in areas that they would like to learn about.

SynthLuvr:
Your project, your rules. However, as an outsider looking in, some consistency would be beneficial. If you want to forbid a group of languages, make the guidelines more clear. On the website it says "Community-driven!" yet when browsing through these discussions it seems a bit contradictory when the community offers additions but they're rejected by the project maintainers based on feelings. This could be solved by adding more guidelines so decisions can be made more objectively.

Issue #4184, opened 2021-06-26
Split listing into "Active" and "Archive"


ad-si:
I was just browsing through the languages and a few of them are basically dead and I think this should be communicated. E.g. whip hasn't seen any development in 8 years. My suggestion would be to split the landing page in 2 sections "Active" and "Archive", where only the languages with changes in the last year (2 years ?) or are known to be still actively maintained are listed under Active.

What do you think?

verhovsky:
I've deleted all the dead languages I've found except for CoffeeScript. If there's more let me know.

Issue #2283, opened 2016-06-21
Found this confusing


proehlen:
Thanks for this reference material - it's very useful.  As a newbie, I found this confusing:
```-- Use case statements to pattern match on different possibilities.
case aList of
  [] -> "matches the empty list"
  [x]-> "matches a list of exactly one item, " ++ toString x
  x::xs -> "matches a list of at least one item whose head is " ++ toString x```

Specifically, the following language:

> matches a list of at least one item whose head is

For me at least, the "whose head is" seemed to indicate matching was conditional on the contents of some pre-populated variable called x.  

I had to play around in code to confirm that the match is just in fact "matches a list of at least one item"

Maybe the sentence could be changed to "matches a list of at least one item.  the head is "?

kok321:
#4949

PR #4949, opened 2024-05-19
[whip/*] delete

verhovsky:
Whip is a Lisp dialect that has 11 GitHub stars and its [last commit](https://github.com/L8D/whip/commits/master/) was in 2013. @L8D 

See #4184


kok321 (commented):
#2283

PR #2701, opened 2017-04-09
[java/fa]add java for Persian Language

ghaseminya:
- [ ] I solemnly swear that this is all original content of which I am the original author
- [x] Pull request title is prepended with `[language/lang-code]`
- [x] Pull request touches only one file (or a set of logically related files with similar changes made)
- [ ] Content changes are aimed at *intermediate to experienced programmers* (this is a poor format for explaining fundamental programming concepts)
- [x] If you've changed any part of the YAML Frontmatter, make sure it is formatted according to [CONTRIBUTING.md](https://github.com/adambard/learnxinyminutes-docs/blob/master/CONTRIBUTING.markdown)
  - [x] Yes, I have double-checked quotes and field names!

PR #4967, opened 2024-06-03
[java/fa] delete

verhovsky:
because it's not actually translated. @ghaseminya #2701


vendethiel:
Oh wow. Thanks for the deletion.

PR #4993, opened 2024-07-04
[python/en] Correct integer division examples (fixes #4989)

### LINKED ISSUE #4989: Python - Statement regarding Integer Division ambiguous
**# Integer division rounds towards zero for both positive and negative numbers.**
5 // 3       # => 1
-5 // 3      # => -2

By statement logic, -5 // 3 should be -1 instead it is rounding off away from zero (towards negative infinity). 
This is by nature is floor division instead of integer division and rounding off towards zero.

Suggest to make suitable change to statement to avoid any possible confusion.

Some research lead towards useful articles mentioned below-

https://python-history.blogspot.com/2010/08/why-pythons-integer-division-floors.html

https://docs.python.org/3/reference/expressions.html?ref=blog.teclado.com#binary-arithmetic-operations

> The / (division) and // (floor division) operators yield the quotient of their arguments. The numeric arguments are first converted to a common type. Division of integers yields a float, while floor division of integers results in an integer; the result is that of mathematical division with the ‘floor’ function applied to the result.

tomaszn:


verhovsky:
Thank you, I confused it with `int(x / y)`.

PR #5007, opened 2024-07-23
[kotlin/en] lambda functions wrong defined

Risuntsy:
lambda functions can not defined by `fun` directly, use `val` (or `var` if mutability is required) instead
```
fun fooLambda: (Int) -> Int = { n -> n + 1 } // Incorrect     X

val fooLambda: (Int) -> Int = { n -> n + 1 } // Correct       ✓
```


vendethiel (approved):

vendethiel:
Thanks!

PR #5068, opened 2024-08-30
[c/de-de] Fix typos

jk4e:


verhovsky:
why are german translators so bad at spelling

PR #5148, opened 2024-10-20
Fix name case

verhovsky:


vendethiel (approved):
Did not check all the files but I agree on principle

PR #5149, opened 2024-10-20
Delete MontiLang

verhovsky:
https://github.com/leowhitehead/MontiLang has 11 GitHub stars and hasn't had a commit in 6 years


verhovsky:
@leowhitehead

leowhitehead:
I've got no objections to this. It was a student project that got some attention at the time, but hasn't really been touched or looked at since.

PR #5306, opened 2025-05-08
[powershell/en] remove comparison operators with bool

awelotta:
- [✅] I solemnly swear that this is all original content of which I am the original author
- [✅] Pull request title is prepended with `[language/lang-code]` (example `[python/fr]` for Python in French or `[java]` for multiple Java translations)
- [✅] Pull request touches only one file (or a set of logically related files with similar changes made)
- [✅] Content changes are aimed at *intermediate to experienced programmers* (this is a poor format for explaining fundamental programming concepts)
- [✅] If you've changed any part of the YAML Frontmatter, make sure it is formatted according to [CONTRIBUTING.md](https://github.com/adambard/learnxinyminutes-docs/blob/master/CONTRIBUTING.md)

### DIFF
--- a/powershell.md
+++ b/powershell.md
@@ -82,12 +82,6 @@ $True * 8    # => '[System.Boolean] * [System.Int32]' is undefined
 [int]$True * 8 # => 8
 $False - 5   # => -5
 
-# Comparison operators look at the numerical value of True and False.
-0 -eq $False  # => True
-1 -eq $True   # => True
-2 -eq $True   # => False
--5 -ne $False # => True
-
 # Using boolean logical operators on ints casts to booleans for evaluation.
 # but their non-cast value is returned
 # Don't mix up with bool(ints) and bitwise -band/-bor


awelotta:
Rationale: the current explanation is not accurate. Rather, the right-hand side converts to the type of the left-hand side, so if the boolean is on the right and a numeric is on the right, we get different results.
```ps1
$False -eq  0 # => True
$True  -eq  1 # => True
$True  -eq  2 # => True
$False -ne -5 # => True
```
I removed it because it seemed like a minor edge-case, but someone who actually has experience programming in PowerShell might have a better idea (would this come up a lot in PS scripts?).

[learn.microsoft on about_Comparison_Operators](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_comparison_operators?view=powershell-7.5#:~:text=The%20equality%20operator%20can%20compare%20objects%20of%20different%20types.%20It%27s%20important%20to%20understand%20that%20the%20value%20on%20the%20right%2Dhand%20side%20of%20the%20comparison%20can%20be%20converted%20to%20the%20type%20of%20the%20left%2Dhand%20side%20value%20for%20comparison.)

PR #5465, opened 2026-01-15
[bash/en]: Multiple instance substitution in variables

KaoKsn:
Add a line that performs multiple instance substitution in a given variable.
Specifically after **Line 77**
```bash
# String substitution in variables:
echo "${variable/Some/A}" # => A string
# This will substitute the first occurrence of "Some" with "A".
```
Bash command added:
```bash
echo "${variable//Some/A}"
```


vendethiel:
Good idea, thanks

PR #5483, opened 2026-03-17
Add ANSI escape codes tutorial

yuxi-liu-wired:
Adds a tutorial for ANSI escape codes, the character sequences that control terminal text formatting, color, and cursor movement.

The tutorial covers:

- ESC character representations (`\e`, `\033`, `\x1b`) and CSI/SGR fundamentals
- Text formatting attributes: bold, dim, italic, underline, blink, inverse, hidden, strikethrough
- Standard 8 colors (foreground 30-37, background 40-47)
- Bright/high-intensity colors (90-97, 100-107)
- 256-color mode (`38;5;N` / `48;5;N`) with the 6x6x6 color cube and grayscale ramp
- 24-bit true color (`38;2;R;G;B` / `48;2;R;G;B`) with a gradient demo
- Cursor movement (up/down/forward/back, absolute positioning)
- Screen and line clearing (erase in display / erase in line)
- Best practices: reset after use, variables for readability, terminal detection, `NO_COLOR`, `tput`

Every example in the code block produces visible output when run with `bash LearnANSI.sh`.

### DIFF
--- a/djot.md
+++ b/djot.md
@@ -9,204 +9,206 @@ filename: learndjot.dj
 derived from CommonMark.
 It aims to be simple to parse, have little rules for edge cases, consistent,
 hard-wrap friendly, and be expressive.
-    
-    Paragraphs are like Markdown.
-    
-    Start a new paragraph.
-    
-    # Headings like markdown
-    
-    ## 2nd level
-    
-    ### 3rd level
-    
-    Setext headings don't exist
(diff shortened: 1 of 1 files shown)


yuxi-liu-wired:
Splitting into two separate PRs: one for the djot lint fix, one for the ANSI tutorial.

PR #5609, opened 2026-09-13
[scheme/en] improving the "common patterns" section and adding a table showing the differences between scheme implementations

xenodesire:
improving the "common patterns" section and adding a table showing the differences between scheme implementations and fixing a conceptual error in the implementation of lazy evaluation

- [x] Pull request title is prepended with `[language/lang-code]` (example `[python/fr]` for Python in French or `[java]` for multiple Java translations)
- [x] If you've changed any part of the YAML Frontmatter, make sure it is formatted according to [CONTRIBUTING.md](https://github.com/adambard/learnxinyminutes-docs/blob/master/CONTRIBUTING.md)


xenodesire:
After reviewing other tutorials—which seem more comprehensive—I realize I kept mine concise because I assumed readers would understand it was a tutorial rather than a full-blown book covering every minute detail. I am now considering whether my approach is sufficient or if you would prefer I include more details regarding Scheme's nuances—which would require greater reader attention—or continue keeping it concise and less descriptive.

Thank you in advance for the opportunity to contribute.

vendethiel:
There's no set rule, honestly. I think it's better to keep the beginning surface-level, and then expand if necessary.
A counterexample would be the Raku (ex Perl 6) that I initially wrote (before samcv rewrote and improved most of it) and if anything, I think it was a bit too long and too windy. YMMV.

xenodesire:
> There's no set rule, honestly. I think it's better to keep the beginning surface-level, and then expand if necessary. A counterexample would be the Raku (ex Perl 6) that I initially wrote (before samcv rewrote and improved most of it) and if anything, I think it was a bit too long and too windy. YMMV.

I’ve also noticed that there are already tutorials covering other implementations. Since I’m focusing specifically on GNU Guile, I’ve been wondering if it might be more beneficial to be less general and more specific—changing the title to "Guile" instead of just "Scheme," given the sheer number of implementations—or if it would be better to consolidate the other tutorials into this Scheme guide, while crediting the original authors alongside myself. Of course, that would depend on the other authors agreeing to it.

xenodesire:
I am marking this as a draft because external demands are limiting my time and I want to make significant changes to the text; I will resolve any conflicts later, at which point it will be ready to merge. For now, however, I see several points that could—or should—be improved.

